Issue #25, opened 2017-05-02
PMVector #tensorProduct symmetric case returns same matrix for all products


apocalypsemystic:
The PMVector #tensorProduct: implementation checks whether the products are #== so that it can return a PMSymmetricMatrix, but the returned matrix does not actually look at the values in the vector; it returns the same PMSymmetricMatrix for all values.

zeros := PMVector zeros: 2.
zeros tensorProduct: (PMVector zeros: 2).
 "a PMVector(0 0)
a PMVector(0 0)"
zeros tensorProduct: zeros.
 "a PMVector(1 2)
a PMVector(2 4)"

ones := PMVector ones: 2.
ones tensorProduct: (PMVector ones: 2).
 "a PMVector(1 1)
a PMVector(1 1)"
ones tensorProduct: ones.
 "a PMVector(1 2)
a PMVector(2 4)"

olekscode:
I have fixed this issue by making a small change in the tensorProduct: message

```
tensorProduct: aVector
    "Answers the tensor product of the receiver with aVector."
    ^self == aVector "dont use #="
        ifTrue: [PMSymmetricMatrix new: self size function:[:x :y| (self at: x) * (self at: y) ]] 
        ifFalse:[PMMatrix rows: ( self collect: [ :a | aVector collect: [ :b | a * b]])]
```

I've made a pull request, but it seems that it contains too many files. Did I do something wrong?

SergeStinckwich:
Sorry Oleks, but we are not using github for PolyMath code at the moment.
You have to commit your code on SmalltalkHub in the inbox repository and then I will move your code after review: http://smalltalkhub.com/#!/~PolyMath

olekscode:
Should I just commit (Save) it to
http://smalltalkhub.com/mc/PolyMath/PolyMathInbox/main, or do I have to
create a Slice?

If I try to commit it, Monticello wants to add about a hundred different
files from the Math-Matrix package (I assume that's because PolyMathInbox
is not up-to-date with PolyMath).

And if I want to create a slice, it asks me for the issue number. I was
looking for PolyMath issues on FogBugz, but I couldn't find them there.

Also, I think I should create a test for this issue. Should I?

On Thu, May 11, 2017 at 10:21 AM, Serge Stinckwich <notifications@github.com

SergeStinckwich:
[...]

PR #27, opened 2017-05-11
FIX PMVector tensorProduct symmetric case (#25)

olekscode:


SergeStinckwich:
we are using SmalltalkHub until now.

Issue #46, opened 2018-06-25
Add a Baseline for the project 


SergeStinckwich:
Before doing #26, we need a baseline for the project.

SergeStinckwich:
TravisCI configuration should be modified in order to use the baseline instead of configuration.

PR #50, opened 2018-06-28
Convert Configuration to Baseline

jecisc:
Convert configuration to baseline.

Fixes https://github.com/PolyMathOrg/PolyMath/issues/46

### DIFF
--- a/src/.properties
+++ b/src/.properties
@@ -1,0 +1,3 @@
+{
+	#format : #tonel
+}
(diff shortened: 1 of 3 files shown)


jecisc:
This will only work after the migration.

jecisc:
It uses Tonel format. Is that what you wanted?

SergeStinckwich:
Yes Tonel is better :-)

SergeStinckwich:
We have some tests failing at the moment, nothing to worry :-)

jecisc:
I just did a dummy convertion. If there is things to change I'll do it.

PR #73, opened 2018-07-14
DualNumberTest>>testRaisedTo made compatible with pharo7

WernerK:
very simple change


SergeStinckwich:
This is great Werner ! All tests are green on Pharo 6.1 32 bits and 64 bits, and the same for Pharo 7.0 !
We just to fix Appveyor.

Issue #85, opened 2018-12-01
All Test Classes Should Be Labelled Correctly (Part 2)


hemalvarambhia:
There are still some `TestCase` classes in production code packages:
Math-TNSE
Math-Permutation

PR #86, opened 2018-12-01
Issue 85

hemalvarambhia:
This resolves issue #85. As a check, after moving the test cases to their appropriate packages, I re-ran the tests to make sure they were still green.


hemalvarambhia:
I've since discovered that there are two other `TestCase` classes living in production packages: 
jacobi_transform_tests
pmvectoraccumulator

If these need to be moved to more sensible packages, then I am happy to move them as part of this PR rather than a separate one, if that's preferred.

hemalvarambhia:
Something strange has happened so I'm closing tgis and reissuing a new PR.

Issue #91, opened 2019-03-11
add support for diagonal function?


fuhrmanator:
I'm doing some software history (co-change) calculations with matrices using PolyMath and I need to get the ~~`diagonal`~~ `principalDiagonal` of a (~~symmetric~~ square) `PMMatrix` to see total changes for each software class. 

I did not find an obvious way to do this, so I'm probably going to write it as an extension. Is this something you want to have as a pull request? Also, perhaps I missed the obvious?

SergeStinckwich:
I dunno how it happens, but there was a missing ] in your commit : https://github.com/PolyMathOrg/PolyMath/commit/3e8d1eecc0453328db408afa5473927d5928f99d#diff-ae5284ffc9b9802fb493ca318d25b044 
You edit the file by hand ?

Your commit was push to dev branch. Thank you.

fuhrmanator:
My edit by hand of what I thought was just a single file caused a huge
change that I don't understand. Can you check this? The ] appears to have
been lost in this merge. I think what happened was I hand edited a file on
a different branch? I am very sorry and won't try that ever again!

https://github.com/PolyMathOrg/PolyMath/commit/3db369c8e5409d0973c7954dd7bff066ea67a4e4

SergeStinckwich:
Ok there is no problem, Christopher :-) We should be more vigilant when we accept PR and avoid as much as possible to edit by hand. Next time I will spend more time looking at the CI.

PR #92, opened 2019-03-11
Support for principalDiagonal function #91

fuhrmanator:


nicolas-cellier-aka-nice:
Note that Matlab `diag` is well defined for non square matrix (it always starts at upper left for diag index = 0).

It also has the ability to retrieve a sub-diagonal (diag index < 0) or a super diagonal (diag index > 0).

A Matlab-like operation is implemented in Smallapack as `diagonalAt: diagIndex`.

src/Math-Matrix/PMMatrix.class.st
```diff
@@ -588,6 +588,17 @@ PMMatrix >> orthogonalize [
 
 ]
 
+{ #category : #'as yet unclassified' }
+PMMatrix >> principalDiagonal [
+	"https://en.wikipedia.org/wiki/Diagonal#Matrices for definitions"
+	| diag |
+	"Check for square"
```
hemalvarambhia: I would say that the comment is redundant as the code is intention revealing enough.

src/Math-Matrix/PMMatrix.class.st
```diff
@@ -588,6 +588,17 @@ PMMatrix >> orthogonalize [
...
+	self isSquare ifFalse: [ self error: 'Diagonal is not defined for a matrix that is not square.' ].
+	diag := PMVector new: self rows size.
```
hemalvarambhia: Here you could use the `numberOfColumns` message (pretty sure that exists) to be a little more direct. What do you think?

fuhrmanator: I agree with both your comments. However, I'm not sure when I'll have time to make these changes in an image and re-commit them.

SergeStinckwich:
I think we can accept and Hemal you could do a modification after if you want.

hemalvarambhia:
I can make the modification in a later PR, sure. Happy for this to be merged otherwise.

src/Math-Matrix/PMMatrix.class.st
hemalvarambhia: I can make the changes in another PR, no problem at all.

SergeStinckwich:
Thank you Christopher !

Issue #81, opened 2018-08-03
2 PCA implementations that give same results but different from Python scikit-learn implementation ...


SergeStinckwich:
We have 2 implementations of PCA : 
- one based on Jacobi Transformation of the covariance matrix
- another one based on SVD.

They give the same results but the result are different from the one you can find with sci-kit learn in Python: 

```python
import numpy as np
from sklearn.decomposition import PCA
X = np.array([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1], [3, 2]])
pca = PCA(n_components=2)
pca.fit(X)
pca.components_
pca.transform(X)
```
``pca.components_`` returns : 

```python
array([[-0.83849224, -0.54491354],
       [ 0.54491354, -0.83849224]])
```

``pca.transform(X)`` returns:

```python
array([[ 1.38340578,  0.2935787 ],
       [ 2.22189802, -0.25133484],
       [ 3.6053038 ,  0.04224385],
       [-1.38340578, -0.2935787 ],
       [-2.22189802,  0.25133484],
       [-3.6053038 , -0.04224385]])
```

I try to implement a flipsvd method like this one : https://github.com/scikit-learn/scikit-learn/blob/4c65d8e615c9331d37cbb6225c5b67c445a5c959/sklearn/utils/extmath.py#L609
but fails until now. 

Please have a look to tests of ``PMPrincipalComponentAnalyserTest``.

hemalvarambhia:
Hi, @SergeStinckwich I looked into the failing test for the Jacobi Transformation form of PCA a little deeper and noted that **transposing and negating the matrix of eigenvectors** results in the test passing (see screenshot):
![transposeandnegate](https://user-images.githubusercontent.com/2665571/51270525-0e3e7b80-19bd-11e9-83a9-126e1614b6c5.png)

The Scikit Learn library, from what I can see, uses the SVD implementation which involves flipping signs. For this test, then, does it make sense to compare the Jacobi implementation directly with their SVD computed one? Should we instead assert that the negated transpose of the Jacobi implementation equals the SVD one?

hemalvarambhia:
Hi, @SergeStinckwich I looked into the failing test for the Jacobi Transformation form of PCA a little deeper and noted that transposing and negating the matrix of eigenvectors results in the test passing:

![transposeandnegate](https://user-images.githubusercontent.com/2665571/51271201-ce789380-19be-11e9-8ea4-d0b09c693668.png)

[...]

PR #93, opened 2019-03-28
[Issue 81] 2 PCA implementations that give same results but different from Python scikit-learn implementation

hemalvarambhia:
This PR has got the scikit-learn based sign flipping algorithm working. We started by extracting a class that, given U and V, performs the sign flipping. As an acceptance test, we used the U and V matrices computed by scikit-learn. 

In the course of this work we found that for the SVD-based implementation of PCA, the V matrix in PolyMath does not match scikit learn's in terms of the signs; the magnitudes of the elements match though (see comments in #81). As a cross-check we used Wolfram Alpha to perform the same decomposition. Interestingly that does match the matrix yielded by PolyMath. We also checked that 

V<sup>-1</sup>M<sup>T</sup>MV 

yields a diagonalised matrix, which it did.

### DIFF
--- a/src/Math-Core/PMVector.class.st
+++ b/src/Math-Core/PMVector.class.st
@@ -115,8 +115,16 @@ PMVector >> accumulateNegated: aVectorOrAnArray [
 
 { #category : #'as yet unclassified' }
 PMVector >> argMax [
+	| precision |
+	precision := 1.0e-10.
+	^ self argMaxTo: precision
+]
+
+{ #category : #'as yet unclassified' }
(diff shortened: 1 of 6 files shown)


SergeStinckwich:
Thank you very much Hemal for your hard work. My only concern is maybe you should rename the ScikitLearnXXX class with a PM prefix.

hemalvarambhia:
I'm happy to conform to that naming style so I'll issue a PR for that today. Thank you for reviewing so quickly :heart:

PR #107, opened 2019-05-06
[issue-104] Some tests methods fail randomly Part 1: BinomialGenerator Tests

hemalvarambhia:
@SergeStinckwich Here, to achieve the certainty we need, I set the seed to zero and removed the square root, but to be honest doing TDD on Random behaviour/phenomena is a weak point of mine so I'd appreciate some advice on how best to proceed. I'm not sure why we're testing that the floor and ceilings of the numbers generated are within 20% of the mean. To be precise I don't know how we came up with the 20%. I'd like to extract it to an intention revealing temp variable but I don't know what it represents.


SergeStinckwich:
Like you, I don't know the rationale behind these tests. I accept your PR and we try to understand later.

Issue #90, opened 2019-02-20
Class PMVector overrides `==`


khinsen:
`PMVector` is the only class in my Pharo image that overrides `ProtoObject >> ==`, whose comment clearly states that no class should ever override it.

SergeStinckwich:
This is part of DHB original code that was copy-paste in PolyMath, so maybe we have to remove it.

AtharvaKhare:
My understanding is that `==` method checks if both arguments refer to the same object (both have same object pointer). In `PMVector ==`, it is comparing every element of the vector. Is this correct?

I think renaming the it to `isEqualTo` or something similar would be better?

AtharvaKhare:
My understanding is that `==` is used to check if two objects pointers refer to the same location (both objects are same) and it is already defined in `ProtoObject`, there is no reason to override it.

The way `PMVector ==` is implemented, it is checking for equality of contents of the objects, and not the objects itself (or memory location of them).

Better would be renaming it to something like `isEqualTo:`.

Am I correct?

AtharvaKhare:
My understanding is that `==` is used to check if two objects pointers refer to the same location (both objects are same) and it is already defined in `ProtoObject`, there is no reason to override it.

The way `PMVector ==` is implemented, it compares the contents of the object and not the memory location. Renaming it to something like `isEqualTo` (or `=`?) would be better.

Am I correct?

AtharvaKhare:
My understanding is that `==` is used to check if two objects pointers refer to the same location (both objects are same) and it is already defined in `ProtoObject`, there is no reason to override it.

The way `PMVector ==` is implemented, it compares the contents of the object and not the memory location. Renaming it to something like `isEqualTo` (or `=`?) would be better.

Am I correct?

khinsen:
@AtharvaKhare Your analysis is correct, but I'd go farther in the conclusion: overriding `==` is not only useless but potentially dangerous because it breaks the semantics of the operation.

You suggest renaming it, but do you actually see any situation in which it might be useful? Note that `=` is already there (inherited from `SequenceableCollection`), but does something different from `PMVector>>==`.

AtharvaKhare:
@khinsen right, I get it :)

Also, looking at the comment and the syntax `== aNumber`, I think it is intended to apply `==` to every element of the vector, so it is not the same as `=`, which tests for equal contents between objects.
So I think it's use is like this (when rename to something like `isEqualTo`:
```
[...]

PR #108, opened 2019-05-09
Removed == method from PMVector

AtharvaKhare:
The == operator should not be overridden. Removed that block entirely due to it having no real application.

Fixes issue #90.


SergeStinckwich:
BTW, `==`is a method, not an operator :-)

PR #129, opened 2019-05-29
Ability to convert vectors to matrices and vice versa

nikhilpinnaparaju:


hemalvarambhia:
Hi, Nikhil. First question I have is what is driving the need for the functionality you're developing?

src/Math-Core/PMVector.class.st
```diff
@@ -237,6 +237,26 @@ PMVector >> productWithVector: aVector [
...
 
+{ #category : #'as yet unclassified' }
+PMVector >> resizeWithRows: numOfRows Cols: numOfCols [
+	
+	| computedRows rowNum colNum |
+	
+	numOfRows isNil ifFalse: [rowNum := numOfRows ].
+	numOfCols isNil ifFalse: [colNum := numOfCols ].
+	
+	(rowNum isNil and: colNum isNil) ifFalse: [
```
hemalvarambhia: I think here we might extract to an `asMatrix` message. It feels like you're doing that. The advantage is that there are fewer nested if and thus the code is easier to understand.

nikhilpinnaparaju:
Motivation - Softmax is primarily used on vectors but it can be applied to matrices as well in a couple of ways. Either along any of the axis of the matrix or we can flatten the matrix and then apply softmax (on the matrix that is now a vector) and then reshape it back to the original shape.

src/Math-Core/PMVector.class.st
nikhilpinnaparaju: I didn't quite understand what you mean here.

nikhilpinnaparaju: The initial part of the code checks if both dimensions - num of cols and num of rows have been given. For example in numpy we can reshape a (10,1) vector to (5,-1) which means (5,2) matrix. The '-1' tells the program to automaticaly identify what the other dimension is.

hemalvarambhia: I have to admit the code is a bit difficult to follow my motivation is to make it easier to understand, I guess. There are a lot of ifs there. Maybe we start with that.

SergeStinckwich: I have also the same concern about the readability.

hemalvarambhia: For example when will `numRows isNil and: numCols isNil` evaluate to false?

nikhilpinnaparaju: When one of the dimensions is not given

nikhilpinnaparaju:
``` smalltalk
|vect|
vect := #(1 0.5 0.2 3 1 -1 7 3 2 12 13 3) asPMVector .
vect resizeWithRows: 2 Cols: nil
```

nikhilpinnaparaju: the whole `nil` business is to replicate the behavior of -1  with numpy reshaping

hemalvarambhia: I would say here that I strongly disagree with the way numpy are doing it, using -1. To me it feels strange. With the example above, are you in fact transposing the column vector so that it becomes a row vector. There is a distinction of row and column vectors from what I recall.

hemalvarambhia: Maybe we can start by refactoring to guard clauses. Can you start by adding a guard clause when only number of columns is nil.

hemalvarambhia: A final question is what scenario would give rise to either `numberOfRows` and `numberOfCols` being nil?

nikhilpinnaparaju: Should I just drop the whole nil thing? Add a checking clause whether the given array of dimensions is compatible or not and then reshape based on it?

nikhilpinnaparaju: Ans to Final Ques) If the user only knows how many rows he wants to have or only knows how many columns he wants to have.

hemalvarambhia: I think we should drop the nil idea, personally. By expecting numbers it keeps the logic simple. It's a contractual obligation on the client to provide those and forces them to think a bit more :D

khinsen: Given how fundamental vectors and matrices are, it's worth spending some time cleaning it up. Both the API and the implementation.

khinsen: Something else to keep in mind: do we want N-dimensional arrays in PolyMath one day? If yes, we should have an API that scales to more than two dimensions. Everything must be doable without references to "rows" or "columns". In particular, the distinction between "row vectors" and "column vectors" has no useful generalization to higher dimensions.

hemalvarambhia: Are N-dimensional arrays really tensors? Forgive my ignorance here.

src/Math-Core/PMVector.class.st
```diff
@@ -237,6 +237,26 @@ PMVector >> productWithVector: aVector [
...
+	
+	numOfRows isNil ifFalse: [rowNum := numOfRows ].
+	numOfCols isNil ifFalse: [colNum := numOfCols ].
+	
+	(rowNum isNil and: colNum isNil) ifFalse: [ 
+		numOfRows isNil ifTrue: [ rowNum := self size / colNum ] .
+		numOfCols isNil ifTrue: [ colNum := self size / rowNum ] .
+		].
+	
+	self assert: (rowNum * colNum = self size) description: 'Number of Rows not compatible with number of Columns'.
```
hemalvarambhia: I think 'rows' and 'columns' rather than 'Rows'/'Columns' as in this case they're not proper nouns.

src/Math-Core/PMVector.class.st
```diff
@@ -237,6 +237,26 @@ PMVector >> productWithVector: aVector [
 			into: [ :sum :each | n := n + 1. (aVector at: n) * each + sum]
 ]
 
+{ #category : #'as yet unclassified' }
```
hemalvarambhia: Should this categorised as 'transformation'?

src/Math-Core/PMVector.class.st
khinsen: The currently fashionable name is indeed tensors. As a physicist, I prefer to reserve that word for its original meaning (a quantity that respects certain transformation laws under changes of coordinate systems), so I stick to the old-fashioned term "N-dimensional array".

src/Math-Core/PMVector.class.st
```diff
@@ -122,6 +122,18 @@ PMVector >> asPMVector [
...
+PMVector >> checkDimensionalCompatibility: dimensionArray [
+	|prod|
+	prod := 1.
+	
+	dimensionArray do: [ :each | prod := prod * each ].
+	
+	self assert: (self size = prod) description: 'Imcompatible combination of Dimensions provided'.
+	
+	^true
```
hemalvarambhia: I think this is unnecessary as you're not relying on the output in places where you send this message.

Issue #105, opened 2019-05-01
Math-Random package is messy


SergeStinckwich:
The ``Math-Random`` is really messy and need to be cleaned :
- Some classes are subclasses of Random class
- some other classes are ``PMNumberGenerator``, some other from ``PMRandomGenerator``.
- What is the difference between ``PMNumberGenerator`` and ``PMRandomGenerator`` ?
- why some classes end with ``Generator`` other with ``RandomGenerator``, others by ``Kernel`` ?
- ``PMMarsagliaKissRandomKernel``is a subclass of ``Object`` and there is another class with almost the same name: ``PMMarsagliaKissRandom``. Maybe is their relationship ?

``PMLaplaceGenerator`` subclasses Object instead of ``PMRandomGenerator`` ...

hemalvarambhia:
@SergeStinckwich With reference to point 3 (What is the difference between `PMNumberGenerator` and `PMRandomGenerator` ?) The `PMNumberGenerator` is dependent upon `PMRandomGenerator`. Every subclass has a `defaultGeneratorClass` class method who default is a subclass of `PMRandomGenerator` (Parker-Miller's algorithm). I think the class `PMRandomGenerator` should be called PMRandomNumberGeneratingAlgorithm, maybe. I will try and read up a little on the subject before renaming, though.

hemalvarambhia:
On point 5, the class comment and [Wikipedia](https://en.wikipedia.org/wiki/KISS_(algorithm)) say that `PMMarsagliaKissRandom` is a PRNG. On the relationship question, `PMMarsagliaKissRandom` dependency on the kernel.

hemalvarambhia:
So far, having reviewed the production code in Math-Random, the Mersenne Twister class's `nextInteger` message is quite long, a little difficult to understand. As a start we could try to extract some of the blocks to intention-revealing methods. I've found the original paper on the Mersenne Twister algorithm so this could be used for inspiration.

hemalvarambhia:
The scope of the issue is potentially very large. I propose we break the issue up into smaller ones. I'll review the code and create issues accordingly.

PR #134, opened 2019-06-05
[Issue 105] Refactor Math-Random Tests Package

hemalvarambhia:
Issue: #105 
This PR continues refactoring the tests package. Here I have tried to use what I believe is the correct domain language so that the tests serve as documentation. However, this area of Mathematics is not my strong point so it'd be great to get some suggestions from fellow maintainers of method names, specifically whether we should use the word 'mean' or 'expected value'.

### DIFF
--- a/src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
+++ b/src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
@@ -5,7 +5,7 @@ Class {
 }
 
 { #category : #tests }
-PMBinomialGeneratorTest >> generateASampleOfSize: numberOfTrials UsingGenerator: gen [
+PMBinomialGeneratorTest >> generateASampleOfSize: numberOfTrials usingGenerator: gen [
 	| sample |
 	sample := OrderedCollection new.
 	numberOfTrials timesRepeat: [ sample add: gen next ].
@@ -58,7 +58,7 @@ PMBinomialGeneratorTest >> testSampleMeanConvergesToExpectedValue [
 	
 	sample := self
 		generateASampleOfSize: numberOfTrials
-		UsingGenerator: gen.
+		usingGenerator: gen.
 	self assert: sample min > (gen expectedValue * (1 - 0.2)).
 	self assert: sample max < (gen expectedValue * (1 + 0.2)).
 	self assert: (sample average - gen expectedValue) abs < 5
--- a/src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
+++ b/src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
@@ -5,6 +5,14 @@ Class {
 }
 
 { #category : #tests }
+PMBinomialGeneratorTest >> generateASampleOfSize: numberOfTrials UsingGenerator: gen [
+	| sample |
+	sample := OrderedCollection new.
+	numberOfTrials timesRepeat: [ sample add: gen next ].
+	^ sample
+]
+
+{ #category : #tests }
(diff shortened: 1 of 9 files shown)


src/Math-Tests-Random/PMPoissonGeneratorTest.class.st
```diff
@@ -4,6 +4,14 @@ Class {
 	#category : #'Math-Tests-Random'
 }
 
+{ #category : #tests }
+PMPoissonGeneratorTest >> generateASampleOfSize: sampleSize UsingGenerator: gen [
```
khinsen: I'd go for `#generateASampleOfSize:usingGenerator:` (with a lower-case 'u'), which is more in line with what I see elsewhere in Pharo.

src/Math-Tests-Random/PMExponentialGeneratorTest.class.st
```diff
@@ -7,6 +7,14 @@ Class {
 	#category : #'Math-Tests-Random'
 }
 
+{ #category : #tests }
+PMExponentialGeneratorTest >> generateASampleOfSize: sampleSize UsingGenerator: gen [
```
khinsen: That's the same method as in `PMPoissonGeneratorTest`. Can we have this only once? Maybe in a common superclass of all random-number tests?

src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
```diff
@@ -4,6 +4,14 @@ Class {
 	#category : #'Math-Tests-Random'
 }
 
+{ #category : #tests }
+PMBinomialGeneratorTest >> generateASampleOfSize: numberOfTrials UsingGenerator: gen [
```
khinsen: One more copy...

src/Math-Tests-Random/PMExponentialGeneratorTest.class.st
hemalvarambhia: I've actually proposed above moving the duplicate code to a helper class for now. I did think about moving the code to a super class but I tend to prefer composition to inheritance.

src/Math-Tests-Random/PMBinomialGeneratorTest.class.st
```diff
@@ -56,7 +48,7 @@ PMBinomialGeneratorTest >> testSampleMeanConvergesToExpectedValue [
 		numberOfTrials: numberOfTrials
 		probabilityOfSuccess: probabilityOfSuccess.
 	
-	sample := self
+	sample := SampleNumericalData
```
khinsen: Using a helper class is fine with me, but since it's in PolyMath it should have the PM prefix in its name. I'd also prefer to have "random" somewhere in the same, as "sampling" is also used for deterministic approaches (e.g. sampling a continues signal to obtain a discrete time series). Perhaps "PMRandomSampling" or something similar?

(Warning: I am notorious for changing my mind about names frequently. I might dislike my own proposition tomorrow.)

hemalvarambhia: I forgot the PM prefix! I'll make that change.

I think given the context is the random package I'm OK with sampling. With reference to deterministic processes, the meaning seems to be different in that context, much like weight in physics is different from (statistical) weight.

hemalvarambhia: Then the code would like like `PMRandomSample ofSize:1000 usingGenerator: generator`
which I certainly think reads well, @khinsen

Issue #106, opened 2019-05-03
Support for coveralls.io


SergeStinckwich:
Add support for tests coverage in the CI with : http://coveralls.io

SergeStinckwich:
Add support here: https://travis-ci.org/PolyMathOrg/PolyMath but all build fails when I modify .smalltalk-ston to use coverage: https://travis-ci.org/PolyMathOrg/PolyMath/builds/527648159

hemalvarambhia:
I'll look to break this task up on a package by package basis as it's become quite big now.

PR #140, opened 2019-07-17
[Issue 106] Get Math-ODE Covered By Coveralls.io 

hemalvarambhia:
Issue #106 

Here I have got the `Math-ODE` package covered by coverall.io. I tried to list 'Math-O*' originally but some travis jobs failed, running out of memory, so as a small step forward I have hard-coded the package name until we can find a better way forward.


SergeStinckwich:
Thank you Hemal !

PR #142, opened 2019-08-04
Use double dispatch for PMDualNumber

### LINKED ISSUE #141: Use double-dispatch for PMDualNumber
Need to do this for `+`,`*`, `-`, `\`. At the moment, these methods are implemented with type checking and this is not really scalable.

SergeStinckwich:
Fix #141

### DIFF
--- a/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
+++ b/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
@@ -22,8 +22,6 @@ Class {
 
 { #category : #'instance creation' }
 PMDualNumber class >> value: aValue [
-	"for a constant"
-
 	^ (self new value: aValue) eps: 0
 ]
 
--- a/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
+++ b/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
@@ -55,17 +55,14 @@ PMDualNumber >> + aNumber [
 ]
 
 { #category : #arithmetic }
-PMDualNumber >> - aDualNumber [
-^aDualNumber isDualNumber
-	ifFalse:[self class value: value - aDualNumber eps: eps]
-	ifTrue:[self class value: value - aDualNumber value eps: eps - aDualNumber eps]
+PMDualNumber >> - aNumber [
+	^ aNumber negated addDualNumber: self
 ]
 
 { #category : #arithmetic }
-PMDualNumber >> / aDualNumber [
-^aDualNumber isDualNumber
-	ifFalse:[self class value: value / aDualNumber eps: eps /aDualNumber]
-	ifTrue:[self class value: value / aDualNumber value eps:( eps * aDualNumber value - (aDualNumber eps *value))/aDualNumber value squared]
+PMDualNumber >> / aNumber [
+
+	^ aNumber reciprocal multiplyDualNumber: self
 ]
 
 { #category : #testing }
@@ -136,8 +133,8 @@ PMDualNumber >> eps [
 ]
 
 { #category : #accessing }
-PMDualNumber >> eps: anObject [
-eps := anObject
+PMDualNumber >> eps: aNumber [
+	eps := aNumber
 ]
 
 { #category : #comparing }
@@ -157,7 +154,7 @@ PMDualNumber >> exp [
 	^ self class value: e eps: eps * e
 ]
 
-{ #category : #hash }
+{ #category : #comparing }
 PMDualNumber >> hash [
 	^ value hash
 ]
--- a/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
+++ b/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
@@ -82,7 +82,7 @@ PMDualNumber >> abs [
 ]
 
 { #category : #arithmetic }
-PMDualNumber >> abs2 [
+PMDualNumber >> absSquared [
 	^ (self conjugated * self) real
 ]
 
--- a/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
+++ b/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
@@ -1,4 +1,7 @@
 "
+In linear algebra, dual numbers extend the real numbers by adjoining one new element ε with the property ε^2 = 0 (ε is nilpotent).
+See here: https://en.wikipedia.org/wiki/Dual_number^
+
 PMDualNumbers can be used to calculate the first derivative if one  creates them this way:
 	PMDualNumber value: aNumber eps:  derivativeOfaNumber (1 if the derivative with respect to aNumber is calculated, 0 otherwise)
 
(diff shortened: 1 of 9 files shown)


src/Math-AutomaticDifferenciation/PMDualNumber.class.st
```diff
@@ -1,26 +1,30 @@
...
 	],
-	#category : 'Math-AutomaticDifferenciation'
+	#category : #'Math-AutomaticDifferenciation'
 }
 
 { #category : #'instance creation' }
-PMDualNumber class >> value:aValue [
-"for entering a constant"
-^(self new value: aValue) eps: 0 
+PMDualNumber class >> value: aValue [
```
hemalvarambhia: If it's easy to change, perhaps we can rename this message to `constant:`. This would eliminate the need for the comment underneath.

src/Math-Complex/PMComplex.class.st
```diff
@@ -198,6 +198,13 @@ PMComplex >> abs [
 	^ ((real * real) + (imaginary * imaginary)) sqrt
 ]
 
+{ #category : #arithmetic }
+PMComplex >> abs2 [
```
hemalvarambhia: Consider renaming this to `squareModulus` or `absSquared`.

src/Math-Tests-AutomaticDifferenciation/PMDualNumberTest.class.st
```diff
@@ -19,18 +19,16 @@ PMDualNumberTest >> assertEquality: aDualNumber and: anotherDualNumber [
...
 { #category : #running }
 PMDualNumberTest >> setUp [
-zero:=PMDualNumber value: 0 eps: 1.
-zeroc:=PMDualNumber value: 0 eps: 0.
-one:=PMDualNumber value: 1 eps: 1.
-onec:=PMDualNumber value: 1.
-three :=PMDualNumber value: 3.0 eps: 1.
-
+	zero := PMDualNumber value: 0 eps: 1.
+	zeroc := PMDualNumber value: 0 eps: 0.
```
hemalvarambhia: What is `zeroc`

src/Math-AutomaticDifferenciation/PMDualNumber.class.st
```diff
@@ -244,6 +275,11 @@ PMDualNumber >> raisedToInteger: anInteger [
...
 ]
 
+{ #category : #converting }
+PMDualNumber >> real [
+	^ self class value: self value real eps: self eps asComplex real
+]
+
 { #category : #arithmetic }
 PMDualNumber >> reciprocal [
 ^self class value: 1/value  eps: eps /value squared negated
```
hemalvarambhia: If it's safe, perhaps indent this line too.

src/Math-AutomaticDifferenciation/PMDualNumber.class.st
SergeStinckwich: I will remove the constant comment. In fact, here, we just build a dual number where the coefficient of epsilon is 0. Maybe eps: can be rename as epsilon:

src/Math-Complex/PMComplex.class.st
SergeStinckwich: ok I will rename it as absSquared.

src/Math-Tests-AutomaticDifferenciation/PMDualNumberTest.class.st
SergeStinckwich: I guess the c means constant, but I don't like the name constant.

src/Math-AutomaticDifferenciation/PMDualNumber.class.st
SergeStinckwich: what do you mean ?

hemalvarambhia: I mean to correct the spacing so that it looks like
```
PMDualNumber >> reciprocal [
       ^self ...
```

SergeStinckwich: ok

src/Math-AutomaticDifferenciation/PMHyperDualNumber.class.st
```diff
@@ -66,160 +66,195 @@ PMHyperDualNumber >> / aHyperDualNumber [
...
 { #category : #accessing }
 PMHyperDualNumber >> eps1eps2 [
-	^eps1eps2
+	^ eps1eps2
 ]
 
 { #category : #accessing }
 PMHyperDualNumber >> eps1eps2: anEps [
-	eps1eps2 := anEps 
+	eps1eps2 := anEps
```
hemalvarambhia: A thought I had was that instances of `PMHyperDualNumber` should be immutable. I'd prefer we address it now. Do you feel the same? We could make it private to ensure clients never get to use it.

SergeStinckwich: You mean remove ``PMDualNumber>>eps:`` and ``PMHyperDualNumber>>eps1eps2:`` ? Make sense, some clients are using them at the moment.

SergeStinckwich: You mean, put ``PMDualNumber>>eps:`` and ``PMHyperDualNumber>>eps1eps2:`` in a private protocol ?

hemalvarambhia: Yes. I feel so, provided it is not being used by other client code.

SergeStinckwich: This is already used in some PMDualNumber and PMHyperDualNumber methods.

hemalvarambhia: Ah. Fair enough. Perhaps in some documentation we should advise against using this method, then because it allows numbers (value objects) to become mutable.

SergeStinckwich: We could use immutability and define them as beReadOnlyObject.

hemalvarambhia: That'd be perfect! :heart:

SergeStinckwich: I try to convert PMDualNumberand PMHyperDualNumber as classes that generate immutable objects. This is ok with PMDualNumber but a bit a pain for PMHyperDualNumber, because the last class is a subclass of the previous one. Then this is not possible to reuse operations with super, because of the immutability. So it means we have to duplicate code between the two classes. There is a kind of tradeoff with immutability and code reuse.

hemalvarambhia: I'd prefer immutability over code reuse. As I was taught by Sandi Metz, it's easier to recover from code duplicate than from bad abstraction. I think I need to understand what a dual and hyper dual number is. Inheritance shouldn't be applied to re-use code. Could there be a Liskov Substitution problem here?

SergeStinckwich: PMHyperDualNumber is a dual number with 2 epsilons values.
PMDualNumber is a dual number with only one epsilon value.
This is why I think these 2 classes will be supersede later by an implementation of multi-dimensional dual numbers.

PR #150, opened 2019-09-02
Cleanup to load in PharoGs

jgfoster:
Various issues with loading into PharoGs. One issue is using an instance variable name as a method argument or method temp. Another is spacing where a dot comes between two names; GemStone treats this as a special syntax for direct instance variable access (eek!).


SergeStinckwich:
Thank you James for your contribution.
Apparently the CI is red, because you introduce a small typo in the method : ``PMKDTreeTest >> testKDTree1Dimension``. There is a space between t and ree. Can you fix it ?

PR #178, opened 2020-08-21
testHouseholder added

mabdi:
I submit this pull request to suggest a test method `PMVectorTest >> testHouseholder`.

We noticed that the method #householder is never executed by any of the tests in PMVectorTest. Therefore we created a test method which uses two vectors to exercise both branches in line 203 `(x <=0) ifTrue: [x -u] ifFalse:  [0 - s / (x + u)]`. In the former vector `x = -1` forces the ifTrue branch in the latter  vector `x = 1.00001` forces the ifFalse branch. 

Note that these suggestions are adapted from a test amplification tool called SmallAmp (https://github.com/mabdi/small-amp). SmallAmp executes existing tests, sees which parts of the class under test are not covered and then suggests improvements on the test methods.

I hope you will accept this pull request. It would illustrate that SmallAmp makes relevant suggestions.

Mehrdad Abdi.

### DIFF
--- a/src/Math-Tests-Core/PMVectorTest.class.st
+++ b/src/Math-Tests-Core/PMVectorTest.class.st
@@ -75,13 +75,11 @@ PMVectorTest >> testHouseholder [
 	u := #(-1 0 1) asPMVector. "`x <= 0` when x = -1"
 	w := u householder.
 	self
-		assert: w class equals: Array;
 		assert: (w at: 1) equals: 1.7071067811865475;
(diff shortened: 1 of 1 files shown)


SergeStinckwich:
Thank you @mabdi for your contribution ! Can you take care about the remarks of Hemal and we can integrate your test.

mabdi:
Of course, I omitted the type-assertions. Please check again.

SergeStinckwich:
Thank you for increasing our test coverage ;-)

Issue #179, opened 2020-12-02
Move CI to Git Actions


SergeStinckwich:
https://badetitou.github.io/misc/pharo/2020/11/30/Testing-pharo-with-github-actions/

hemalvarambhia:
I'm going to look at @Ducasse 's [prior art](https://github.com/pharo-containers/Containers-OrderPreservingDictionary/blob/master/.github/workflows/matrix.yml)

PR #183, opened 2020-12-30
Add an action to run smalltalkCI on a PR being made or a commit is pushed.

hemalvarambhia:
Issue: #179 

Following the link provided, I have created a YAML file that runs smalltalkCI when a commit is pushed and a pull request is made.

For this PR I have asked for the build to be done on ubuntu only.

Question:
Do we want to continue having builds done on macOS and Windows?


.github/workflows/smalltalk-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the development branch
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
hemalvarambhia, line 12: Do we want to continue testing on macOS and windows? My guess is yes

Ducasse: Yes!

hemalvarambhia: Addressed in [this commit](https://github.com/PolyMathOrg/PolyMath/pull/183/commits/0ff00106d336bc7f2017653a8a5d24f77e3d4f07)

SergeStinckwich (approved):

PR #184, opened 2021-01-01
Add reverse binomial with tests.

hernanmd:
See https://math.stackexchange.com/a/103385/205 for details


src/Math-Numerical/Integer.extension.st
```diff
@@ -1,5 +1,54 @@
 Extension { #name : #Integer }
 
+{ #category : #'*Math-Numerical' }
+Integer >> findNK [
+	" Reverse binomial coefficient. Answer a <Collection> with all n and k such that n take: k = self. Elements in the answered Collection should be read as paired. Each pair represents (n,k) in the binomial coefficient formula. 
+	See https://math.stackexchange.com/a/103385/205 for details. "
+
+	| k |
+	[ self isInteger and: [ self > 1 ] ] assert.
```
hemalvarambhia, line 9: the method is a little long. Is there a way to break it up?

src/Math-Tests-Numerical/IntegerTest.extension.st
```diff
@@ -0,0 +1,42 @@
...
+	self should: [ 1 findNK ] raise: Error.
```
hemalvarambhia, line 15: Move these to test methods as proposed by @olekscode perhaps?

src/Math-Numerical/Integer.extension.st
```diff
@@ -1,5 +1,54 @@
 Extension { #name : #Integer }
 
+{ #category : #'*Math-Numerical' }
+Integer >> findNK [
```
hemalvarambhia, line 4: Maybe `reverseBinomial:` given your comment  in the method block?

src/Math-Numerical/Integer.extension.st
hemalvarambhia, line 9: Given you've defined this in the Integer class, do you need to have the `isInteger` check?

src/Math-Numerical/Integer.extension.st
```diff
@@ -1,5 +1,54 @@
...
+						stream nextPutAll: { n asInteger . k asInteger }.
+						k < (n - k)
+							ifTrue: [ stream nextPutAll: { n asInteger . (n - k) asInteger } ] ].
+	            choose := choose * (n + 1).
+	            choose := (choose / (n + 1 - k)) ceiling ] ] ]
+
+]
+
+{ #category : #'*Math-Numerical' }
+Integer >> firstOver: c [
```
hemalvarambhia, line 33: Maybe add a link to the underlying Mathematics? This is a part of the domain I'm not clear on.

hemalvarambhia:
As a first observation this seems to be a direct translation of the python code snippet in the stack exchange link.

src/Math-Numerical/Integer.extension.st
hernanmd: You're right! I removed the check. Thanks!

hemalvarambhia:
@hernanmd as I mentioned above, the code appears to be python translated, almost literally to Smalltalk. Is this the case? 

I think `findNK` could be made more intention revealing. Could the code be refactored a little as I can't seem to follow it too easily (I'm having an off day!).

hernanmd:
> @hernanmd as I mentioned above, the code appears to be python translated, almost literally, to Smalltalk. Is this the case?
> 
> I think `findNK` could be made more intention revealing. Could the code be refactored a little as I can't seem to follow it too easily (I'm having an off day!).

Yes, it is a translation from Python code that's why I included the link :)

I don't see how it could be refactored because Integer can easily explode.

src/Math-Tests-Numerical/IntegerTest.extension.st
```diff
@@ -0,0 +1,42 @@
...
+	self assert: (3 firstOver: 4) equals: 2.
+	self assert: (3 firstOver: 5) equals: 2.
+	self assert: (3 firstOver: 6) equals: 2.
+
+	self assert: (4 firstOver: 2) equals: 2.
+	self assert: (4 firstOver: 3) equals: 2.
+	self assert: (4 firstOver: 4) equals: 2.
+	self assert: (4 firstOver: 5) equals: 2.
+	
+	self assert: (4 firstOver: SmallInteger maxVal) equals: 32768.
```
nicolas-cellier-aka-nice, line 37: SmallInteger maxVal will vary in 32bits vs 64bits image flavours...
Please use an explicit value

hemalvarambhia:
Chasing up on this PR, @hernanmd . Let us know when you've addressed review comments and then we can merge

hemalvarambhia:
@hernanmd chasing up again :)

src/Math-Numerical/Integer.extension.st
```diff
@@ -1,5 +1,54 @@
...
```
nicolas-cellier-aka-nice, line 33: firstOver: is not really intention revealing...

If I understand, that's solving `n^self>=c`, that is `n=(c^(1/self)) ceiling`, or in other words, that's `(c nthRoot: self) ceiling`.
Both Cuis and Squeak have `nthRoot:`. If Pharo lacks such method, then it shall be added to Pharo.

src/Math-Tests-Numerical/IntegerTest.extension.st
hernanmd: Done, thanks!

src/Math-Numerical/Integer.extension.st
hernanmd: I quickly checked but `2 nthRoot: 3`in Squeak has different output than `2 firstOver: 3`.

SergeStinckwich:
Move to 1.0.4 milestone

SergeStinckwich:
Can we merge? @hemalvarambhia

hemalvarambhia:
@SergeStinckwich I think the code quality may be improved. Plus, there has been no addressing of comments or any other activity. 

I would close it rather than merge. That way, it can reopened when the author has energy and capacity. 

What do other reviewers think?

hernanmd:
> @SergeStinckwich I think the code quality may be improved. Plus, there has been no addressing of comments or any other activity.

Can you detail which aspect of the code could be improved?

IMHO it seems too much revision for a single method addition, I would love to know why or simply discard this addition definitely.

SergeStinckwich:
The main concern of @hemalvarambhia is about names. Could you use better names for ``findNK`` and ``firstOver:`` ?

SergeStinckwich:
I propose to replace ``findNK`` by ``findAllNKGivenBinomialCoeff``or ``inverseBinomialCoefficient``.
You can remove ``firstOver:``and replace directly in the code: ``nmin := k firstOver: (k factorial * self).``by ``nmin :=  (k factorial * self) nthRoot: k.``.

hemalvarambhia:
> I propose to replace ``findNK`` by ``findAllNKGivenBinomialCoeff``or ``inverseBinomialCoefficient``.
> You can remove ``firstOver:`` (and tests) and replace directly in the code: ``nmin := k firstOver: (k factorial * self).``by ``nmin :=  (k factorial * self) nthRoot: k.``

`inverseBinomialCoeffient` feels more concise IMHO. Could you use Kent Beck's 'Refactor to composed method' to improve the maintainability. This is the only concern that I have.

hemalvarambhia:
@hernanmd This is only my opinion and it is important that others weigh in. If more disagree with my view, ignore the following:

my concern is the structure of `findNK`. It feels very nested, with `while` and nested `ifTrue` messages so I sense there is a risk to maintainability and understandability, a key trade off for me. The method is also long. One option to improve things might be to appeal to 'refactor to composed method' (cf Kent Beck's 'Smalltalk Best Practice Pattern')

SergeStinckwich:
Thank you Hernan.

SergeStinckwich (approved):

PR #186, opened 2021-01-06
Create Math-Statistics package and add harmonic and geometric mean, and mode

ctSkennerton:
Hi,

Would you be interested in having a few extra basic statistics measures added into Polymath?

Best,
Connor


src/Math-Statistics/Collection.extension.st
```diff
@@ -0,0 +1,37 @@
...
+
+{ #category : #'*Math-Statistics' }
+Collection >> harmonicMean [
+	"Answer with the harmonic mean of the data."
+	"#(2.5, 3, 10) harmonicMean >>> 3.6"
+	| sum |
+	sum := 0.
+	
+	self do: [ :i | i < 0 
+		ifTrue: [Error new signal: 'The harmonic mean cannot be calculated on negative numbers'] 		ifFalse: [ sum := (sum + (1 / i))]
```
hemalvarambhia, line 21: This could be a guard clause.

src/Math-Statistics/Collection.extension.st
```diff
@@ -0,0 +1,37 @@
+Extension { #name : #Collection }
+
+{ #category : #'*Math-Statistics' }
```
hemalvarambhia, line 3: @SergeStinckwich I suggested that geometric and harmonic mean computation should be in our Statistics package after some quick research on Wikipedia. Would you agree? Are there alternatives?

src/Math-Statistics/Collection.extension.st
ctSkennerton: Refactored this to use a guard clause.

src/Math-Statistics/Collection.extension.st
nicolas-cellier-aka-nice: Note that there is an arithmetic-geometric mean already implemented in ArbitraryPrecisionFloat.
This is absolutely not specific to statistics package.
Eventually, for small utilities like that, we can tolerate duplications, or we may extract those common utilities in a specific package...

nicolas-cellier-aka-nice: Note that the `agm:` is implemented in `ArbitraryPrecisionFloat`, my understanding is that you want this generalized to a collection, right? In this case, there is no conflict, just two implementations.

nicolas-cellier-aka-nice: Oups, case of miss-reading, you don't want a arithmetic-geometric mean, you just want indepedent arithmetic, geometric, harmonic means? Sorry for the noise.

src/Math-Statistics/Collection.extension.st
```diff
@@ -0,0 +1,37 @@
...
+Collection >> geometricMean [
+	"Answer with the geometric mean of the collection"
+	"#(1 1 2 3 4 5 5 6 6 7 8 9) geometricMean  >>> 3.8583980015011208"
+	"{ 4. 1.  1 / 32} geometricMean >>> (1/2)"
+	"#(3.14 1 4.56 0.333) geometricMean >>> 1.4776945822943937"
+
+	^(self reduce: [ :a :b | a * b ] ) raisedToFraction:  1 / self size.
```
nicolas-cellier-aka-nice, line 10: Note that this may result in Float overflow on large collections.
An alternative is to sum the logarithms.   `(a*b*c)^(1/3)=exp(ln(a*b*c))^(1/3)=exp((ln(a)+ln(b)+ln(c))/3)`
That is something like `^((self collect: #ln) sum / self size) exp`.
The price is that we may answer an inexact Float instead of an exact Integer in some edge cases.
For example`{1/3. 2/3. 8/3. 16/3} geometricMean` would answer 1.3333333333333333 instead of (4/3)...
Also note that the geometricMean may not be well defined for negative numbers, likewise the harmonicMean...

ctSkennerton: Thanks for pointing this out. I have updated the implementation with your suggestion and added a guard clause similar to harmonicMean.

hemalvarambhia:
Wondering whether this introduction of geometric and harmonic means should be an extension to the Collection class. Here I plead ignorance. Could an expert enlighten me.

My thinking is we introduce the idea of a `Statistical Sample` that is backed by a collection. Thoughts?

ctSkennerton:
I agree, it only makes sense if the collection contains a numeric type. When I was implementing these I took inspiration from the statistical methods already defined on collections that similarly only work if they contain numbers. I'm new to smalltalk so I don't know what's considered good style in this case.

hemalvarambhia:
@ctSkennerton hi there. Could you try to extract a `Statistical Sample` class as a refactor. I'd love to see its impact on the code

src/Math-Statistics/Collection.extension.st
```diff
@@ -0,0 +1,44 @@
...
+	^((self collect: #ln) sum / self size) exp
+]
+
+{ #category : #'*Math-Statistics' }
+Collection >> harmonicMean [
+	"Answer with the harmonic mean of the data."
+	"#(2.5, 3, 10) harmonicMean >>> 3.6"
+	| sum |
+	
+	self detect: [ :i | i < 0 ] ifFound: [ Error new signal: 'The harmonic mean cannot be calculated on negative numbers' ] .
```
nicolas-cellier-aka-nice, line 26: maybe test `i <= 0` here too

src/Math-Statistics/Collection.extension.st
```diff
@@ -0,0 +1,44 @@
...
+
+{ #category : #'*Math-Statistics' }
+Collection >> geometricMean [
+	"Answer with the geometric mean of the collection"
+	"#(1 1 2 3 4 5 5 6 6 7 8 9) geometricMean  >>> 3.8583980015011217"
+	"{ 4. 1.  1 / 32} geometricMean >>> 0.49999999999999994"
+	"#(3.14 1 4.56 0.333) geometricMean >>> 1.4776945822943937"
+	"{1/3. 2/3. 8/3. 16/3} geometricMean >>> 1.3333333333333335"
+
+	self detect: [ :i | i < 0 ] ifFound: [ Error new signal: 'The geometric mean should not be calculated on negative numbers' ] .
```
nicolas-cellier-aka-nice, line 11: maybe test `i <= 0`

ctSkennerton:
@hemalvarambhia Maybe I can sketch out what a `StatisticalSample` class would be to see if I understand correctly. I'm thinking there would have a number of class methods for the various operations that take a collection as an input parameter.

```smalltalk
StatisticalSample class >>> harmonicMean: aCollection
StatisticalSample class >>> geometricMean: aCollection
StatisticalSample class >>> mode: aCollection
```

(This could then be further developed with other measures as needed)

What do you think?

hemalvarambhia:
> @hemalvarambhia Maybe I can sketch out what a `StatisticalSample` class would be to see if I understand correctly. I'm thinking there would have a number of class methods for the various operations that take a collection as an input parameter.
> 
> ```smalltalk
> StatisticalSample class >>> harmonicMean: aCollection
> StatisticalSample class >>> geometricMean: aCollection
> StatisticalSample class >>> mode: aCollection
> ```
> 
> (This could then be further developed with other measures as needed)
> 
> What do you think?

You've got the idea. My thinking is that the methods would be instance ones, with the collection being a state variable.

ctSkennerton:
Ok, I've converted the code to use a new class `PMStatisticalSample`

src/Math-Statistics/Collection.extension.st
ctSkennerton: Thank you, I've changed the logic to be `<=0`

hemalvarambhia (approved):

hemalvarambhia:
Merge at your leisure

ctSkennerton:
@hemalvarambhia I don't have permission to merge.

hemalvarambhia:
@ctSkennerton merged. Thank you for contributing. Looking forward to your next one. :)

hemalvarambhia:
@ctSkennerton love the documentation by the way. Very thorough!

PR #189, opened 2021-02-10
Add one example of use of PMHillClimbingOptimizer

SergeStinckwich:
Add start to wrote class comment for this class.
Clean the code

### DIFF
--- a/src/Math-Numerical/PMHillClimbingOptimizer.class.st
+++ b/src/Math-Numerical/PMHillClimbingOptimizer.class.st
@@ -1,3 +1,9 @@
+"
+The class PMHillClimbingOptimizer is a subclass of class PMFunctionOptimizer.
+
+It has only one additional instance variable, unidimensionalFinder to hold the one-dimensional
+optimizer used to find an optimum of the goal function along a given direction.
+"
 Class {
 	#name : #PMHillClimbingOptimizer,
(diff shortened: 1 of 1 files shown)


src/Math-Numerical/PMHillClimbingOptimizer.class.st
```diff
@@ -7,6 +13,21 @@ Class {
 	#category : #'Math-Numerical-Math-FunctionIterator'
 }
 
+{ #category : #examples }
+PMHillClimbingOptimizer class >> example1 [
```
hemalvarambhia, line 17: Maybe this could be more intention revealing? In what way is this special?

src/Math-Numerical/PMHillClimbingOptimizer.class.st
```diff
@@ -1,3 +1,9 @@
+"
+The class PMHillClimbingOptimizer is a subclass of class PMFunctionOptimizer.
```
hemalvarambhia, line 2: I think this is clear from the code itself so this, I feel, may be removed.

hemalvarambhia (commented):
Some non blocking comments from me. Let me know on Discord when you've addressed the comments (feel free to disagree) and then I will approve.

SergeStinckwich:
try to address your comments.

hemalvarambhia (approved):

PR #190, opened 2021-02-23
Some basic cleaning on various Distributions classes

SergeStinckwich:


src/Math-Numerical/PMFisherTippettDistribution.class.st
```diff
@@ -68,12 +68,12 @@ PMFisherTippettDistribution >> distributionValue: aNumber [
 
 { #category : #initialization }
 PMFisherTippettDistribution >> initialize: aNumber1 scale: aNumber2 [
-		"Private - Initialize the parameters of the receiver."
+	"Initialize the parameters of the receiver."
+
 	aNumber2 > 0
-		ifFalse: [ self error: 'Illegal distribution parameters'].
+		ifFalse: [ self error: 'Illegal distribution parameters' ].
```
hemalvarambhia, line 74: Here, for me, it would be clearer if we could invert the clause and say

`number < 0 ifTrue: [...]`

src/Math-Core-Distribution/PMNormalDistribution.class.st
```diff
@@ -41,20 +41,19 @@ PMNormalDistribution class >> new: aNumber1 sigma: aNumber2 [
...
-				]
-		ifFalse:[ v1 :=NextRandom.
-				  NextRandom := nil.
-				].
-	^v1
+	NextRandom
+		ifNil: [ [ v1 := Number random * 2 - 1.
+			v2 := Number random * 2 - 1.
+			w := v1 squared + v2 squared.
+			w > 1 ] whileTrue: [  ].
```
hemalvarambhia, line 50: Could you help me understand why the block is empty here?

hemalvarambhia (commented):
The build is red here. Do we need to add some missing programmer tests?

src/Math-Core-Distribution/PMNormalDistribution.class.st
```diff
@@ -41,20 +41,19 @@ PMNormalDistribution class >> new: aNumber1 sigma: aNumber2 [
...
-				  y := ( ( w ln * 2 negated) / w) sqrt.
-				v1 := y * v1.
-				NextRandom := y * v2.
-				]
-		ifFalse:[ v1 :=NextRandom.
-				  NextRandom := nil.
-				].
-	^v1
+	NextRandom
+		ifNil: [ [ v1 := Number random * 2 - 1.
```
hemalvarambhia, line 47: This part is a little hard for me to follow to be honest.  If you could factor out this block to an intention revealing method to begin which, it would be very helpful.

src/Math-Numerical/PMFisherTippettDistribution.class.st
SergeStinckwich: It would be ``aNumber2 <=0 ifTrue: [...]`` I guess.
Why clearer?

SergeStinckwich:
The build is red, because the coverage decreased I guess.

src/Math-Core-Distribution/PMNormalDistribution.class.st
SergeStinckwich: I dunno, I can remove it.

src/Math-Core-Distribution/PMNormalDistribution.class.st
SergeStinckwich: I have no idea what the block is doing ...

nicolas-cellier-aka-nice: What? It's well known Box-Muller algorithm for transforming a uniform distribution into a normal distribution.
See https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform

We need a pair (v1,v2) inside the unit circle.
If it's not, we have to redraw another pair.
Just replace `whileTrue: [ ].` with `whileTrue.`

SergeStinckwich: I replace ``whileTrue:[]`` with ``whileTrue``. I add a comment for the moment to say that we are using the Box-Muller transform. Do you see a better way to be more intention revealing @hemalvarambhia ?

hemalvarambhia: You could extract act boxMullerAlgorithm method

SergeStinckwich:
I do the extraction as boxMullerTransform.

hemalvarambhia (approved):
Much better.

src/Math-Core-Distribution/PMNormalDistribution.class.st
```diff
@@ -41,20 +54,15 @@ PMNormalDistribution class >> new: aNumber1 sigma: aNumber2 [
...
-				v1 := y * v1.
-				NextRandom := y * v2.
-				]
-		ifFalse:[ v1 :=NextRandom.
-				  NextRandom := nil.
-				].
-	^v1
+
+	| v1  |
+	NextRandom
```
hemalvarambhia, line 59: I wondered what kind of variable this is. It seems global. Is that right?

src/Math-Core-Distribution/PMNormalDistribution.class.st
```diff
@@ -8,9 +8,22 @@ Class {
 	#classVars : [
 		'NextRandom'
 	],
-	#category : 'Math-Core-Distribution'
+	#category : #'Math-Core-Distribution'
 }
 
+{ #category : #information }
+PMNormalDistribution class >> boxMullerTransform [
```
hemalvarambhia, line 15: What's interesting is that this is a class method, rather than an instance one. I presume this choice was made because NextRandom is a class variable?

src/Math-Core-Distribution/PMNormalDistribution.class.st
SergeStinckwich: no this is a class variable.

src/Math-Core-Distribution/PMNormalDistribution.class.st
SergeStinckwich: no because ``PMNormalDistribution class >> random`` is a class method.

PR #195, opened 2021-02-24
Implemented Cholesky Decomposition and the Multivariate Normal Distribution

olekscode:
This pull request includes the following contributions:

### Added Number >> isRealNumber

```Smalltalk
5 isRealNumber.  "true"
(5 + 2i) isRealNumber. "false"
```

### Added  isComplexConjugateOf:

```Smalltalk
5 isComplexConjugateOf: 5. "true"
(5 + 2i) isComplexConjugateOf: (5 - 2i). "true"
5 isComplexConjugateOf: (5 + 0i). "true"
```

### Added PMVector >> isReal and PMMatrix >> isReal

Answer true if and only if all values of vector or matrix are real numbers

```Smalltalk
a := #(1 4 -2 0.5) asPMVector.
b := { 1 . 4 . -2 . 0.5 + 3i } asPMVector.

a isReal. "true"
b isReal. "false"
```

```Smalltalk
a := PMMatrix rows: #( 
    (1 3 0)
    (1 2 -4)
    (-8 5 -1)).

b := PMMatrix rows: { 
    { 1 . 3 + 4i . 0 } .
    { 1 - 2i . 2 . -4 } .
    { -8 + 0i . 5 . -1 } }.

a isReal. "true"
b isReal. "false"
```

### Added PMMatrix >> isHermitian, isPositiveDefinite, isNegativeDefinite, isPositiveSemiDefinite, and isNegativeSemiDefinite

The matrix is:

- **Hermitian** if it is square and the element in the i-th row and j-th column is equal to the complex conjugate of the element in the j-th row and i-th column, for all indices i and j.
- **Positive-definite** if it is Hermitian and all of its eigenvalues are positive
- **Negative-definite** if it is Hermitian and all of its eigenvalues are negative
- **Positive semi-definite** if it is Hermitian and all of its eigenvalues are non-negative
- **Negative semi-definite** if it is Hermitian and all of its eigenvalues are non-positive

```Smalltalk
matrix isHermitian.
matrix isPositiveDefinite.
matrix isNegativeDefinite.
matrix isPositiveSemiDefinite.
matrix isNegativeSemiDefinite.
```

### Implemented Cholesky decomposition

[Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition) is a decomposition of a Hermitian, positive-definite matrix into the product of an upper triangular matrix and its conjugate transpose.

```Smalltalk
matrix := PMMatrix rows: #(
    (4 12 -16)
    (12 37 -43)
    (-16 -43 98)).

upperTriangular matrix choleskyDecomposition. "PMMatrix rows: #(
    (2 6 -8)
    (0 1 5)
    (0 0 3))."

(upperTriangular transpose * upperTriangular) closeTo: matrix. "true"
```

### Implemented PMMultivariateNormalDistribution

(based on the Cholesky decomposition of a covariance matrix)

```Smalltalk
mu := #(1 0) asPMVector.

sigma := PMMatrix rows: #(
    (1 0.8)
    (0.8 1)).

distribution := PMMultivariateNormalDistribution
    meanVector: mu
    covarianceMatrix: sigma.

"Random value sampled from a bivariate distribution"
distribution random. "a PMVector(2.604841992662223 1.0531036953728659)"

"Probability of a given value according to the distribution"
distribution value: #(-1 1) asPMVector. "3.002875190198519e-6"
```

Another example:

```Smalltalk
numberOfPoints := 10000.
scale := 30.
pointExtent := 2@2.

color1 := Color fromString: '3288BD'.
color2 := Color fromString: 'D53E4F'.

mu1 := #(1 0) asPMVector.
mu2 := #(-1.5 -1) asPMVector.

sigma1 := PMMatrix rows: #(
	(0.5 1)
	(1 3)).
	
sigma2 := PMMatrix rows: #(
	(1 0.8)
	(0.8 1)).

dist1 := PMMultivariateNormalDistribution
	meanVector: mu1
	covarianceMatrix: sigma1.
	
dist2 := PMMultivariateNormalDistribution
	meanVector: mu2
	covarianceMatrix: sigma2.
	
data1 := (1 to: numberOfPoints) collect: [ :i | dist1 random ].
data2 := (1 to: numberOfPoints) collect: [ :i | dist2 random ].

canvas := RSCanvas new.

data1 do: [ :each |	
	canvas add: (RSEllipse new
		extent: pointExtent;
		x: each first * scale;
		y: each second * scale;
		color: color1;
		yourself) ].

data2 do: [ :each |	
	canvas add: (RSEllipse new
		extent: pointExtent;
		x: each first * scale;
		y: each second * scale;
		color: color2;
		yourself) ].

canvas open
```

Screenshot 2021-02-24 at 01 04 22

### DIFF
--- a/src/Math-Complex/Object.extension.st
+++ b/src/Math-Complex/Object.extension.st
@@ -7,3 +7,10 @@ Object >> isComplexNumber [
 	^ false
 
 ]
+
+{ #category : #'*Math-Complex' }
+Object >> isRealNumber [
+	"Answer true if receiver is a real number. False by default."
+	^ false
+
+]
--- a/src/Math-Core-Distribution/PMMultivariateNormalDistribution.class.st
+++ b/src/Math-Core-Distribution/PMMultivariateNormalDistribution.class.st
@@ -1,0 +1,98 @@
+Class {
+	#name : #PMMultivariateNormalDistribution,
+	#superclass : #PMProbabilityDensity,
+	#instVars : [
+		'meanVector',
+		'covarianceMatrix'
+	],
+	#category : #'Math-Core-Distribution'
+}
+
+{ #category : #'instance creation' }
+PMMultivariateNormalDistribution class >> meanVector: aMeanVector covarianceMatrix: aCovarianceMatrix [
+	^ self new
+		initializeMeanVector: aMeanVector
+		covarianceMatrix: aCovarianceMatrix;
+		yourself
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> average [
+	^ self meanVector
+]
+
+{ #category : #transformation }
+PMMultivariateNormalDistribution >> changeParametersBy: aVector [
+	"Modify the parameters of the receiver by aVector."
+	meanVector := meanVector + aVector first.
+	covarianceMatrix := covarianceMatrix + aVector second.
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> covarianceMatrix [
+	^ covarianceMatrix 
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> distributionValue: aNumber [
+	"Answers the probability of observing a random variable distributed according to the receiver with a value lower than or equal to aNumber. Also known as the cumulative density function (CDF)."
+	
+	self shouldBeImplemented 
+]
+
+{ #category : #initialization }
+PMMultivariateNormalDistribution >> initializeMeanVector: aMeanVector covarianceMatrix: aCovarianceMatrix [
+	meanVector := aMeanVector.
+	covarianceMatrix := aCovarianceMatrix.
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> kurtosis [
+	"Kurtosis is a measure of the 'tailedness' of the probability distribution of a real-valued random variable. Not defined for a multivariate normal distribution"
+	self shouldNotImplement
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> meanVector [
+	^ meanVector 
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> parameters [
+	"Returns an Array containing the parameters of the distribution. 
+	It is used to print out the distribution and for fitting."
+
+	^ { meanVector . covarianceMatrix }
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> power [
+	"Number of dimensions of a multivariate normal distribution"
+	^ meanVector size
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> random [ 
+	"Answer a vector of random numbers distributed accroding to the receiver."
+	| standardNormalRandom upperTriangular |
+	
+	standardNormalRandom := (1 to: self power) asPMVector
+		collect: [ :each | PMNormalDistribution random ].
+		
+	upperTriangular := covarianceMatrix choleskyDecomposition.
+	
+	^ upperTriangular transpose * standardNormalRandom + meanVector
+]
+
+{ #category : #information }
+PMMultivariateNormalDistribution >> skewness [
+	"Skewness is a measure of the asymmetry of the probability distribution of a real-valued random variable about its mean. Not defined for a multivariate normal distribution"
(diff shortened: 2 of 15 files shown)


SergeStinckwich:
There is some conflict @olekscode can you resolve them?

src/Math-Complex/Object.extension.st
```diff
@@ -7,3 +7,10 @@ Object >> isComplexNumber [
 	^ false
 
 ]
+
+{ #category : #'*Math-Complex' }
+Object >> isRealNumber [
+	"Answer true if receiver is a real number. False by default."
+	^ false
+
```
SergeStinckwich, line 15: ``isRealNumber`` or ``isFloatNumber`` ?

src/Math-Core-Distribution/PMMultivariateNormalDistribution.class.st
```diff
@@ -0,0 +1,98 @@
+Class {
+	#name : #PMMultivariateNormalDistribution,
+	#superclass : #PMProbabilityDensity,
```
SergeStinckwich, line 3: is a multivariate normal distribution also a distribution? I'm not sure for example that ``kurtosis`` or ``skewness`` apply in this context.

olekscode:
Hi @SergeStinckwich 

I have fixed the failing tests. Let's see if the build passes this time.

Can we consider multivariate distributions as "distributions" in the same sense as univariate distributions? I don't know.
I think that multivariate distributions are a generalisation to random variables with more than one dimension
And certain statistical concepts such as skewness and kurtosis are only defined for one dimension.
But I see no reason why we can not define those things for multiple dimensions

`isRealNumber` or `isFloatNumber`? I can try to answer:
"Float number" is a computer science term (based on how they are treated by the compiler).
"Real number" is a mathematical term (subset of complex numbers a + bi where b = 0)
5 is a real number but not a float number.
5 is integer, 5.0 is float but in mathematical sense, 5 = 5.0 є R

The visualization depends on Roassal. 
Roassal is included in Pharo 9 but not in Pharo 8
Do you think I should still add the example to the class side? Because that would add dependency on Roassal

SergeStinckwich:
> I have fixed the failing tests. Let's see if the build passes this time.

yes they pass.

> Can we consider multivariate distributions as "distributions" in the same sense as univariate distributions? I don't know.
> I think that multivariate distributions are a generalisation to random variables with more than one dimension
> And certain statistical concepts such as skewness and kurtosis are only defined for one dimension.
> But I see no reason why we can not define those things for multiple dimensions

Maybe for the moment, let's univariate distributions and multivariate distributions.

> `isRealNumber` or `isFloatNumber`? I can try to answer:
> "Float number" is a computer science term (based on how they are treated by the compiler).
> "Real number" is a mathematical term (subset of complex numbers a + bi where b = 0)
> 5 is a real number but not a float number.
> 5 is integer, 5.0 is float but in mathematical sense, 5 = 5.0 є R

I'm not sure, I understand, because there is no real number in a computer, we only represent Float ...
5 is an integer.
From what I understood, you want to separate complex numbers from others?


> The visualization depends on Roassal.
> Roassal is included in Pharo 9 but not in Pharo 8
> Do you think I should still add the example to the class side? Because that would add dependency on Roassal

No, we will not add the Roassal dependency to PM for Pharo 8, but would be nice to have the example for Pharo 9.

olekscode:
> From what I understood, you want to separate complex numbers from others?
Yes

Certain algorithms in algebra are defined only for real matrices (the ones that do not have complex numbers)
For example, Cholesky decomposition can only be applied to Hermitian positive-definite matrices.

Hermitian matrix is a generalization of symmetry to complex matrices:
- if matrix is real (filled with real numbers) then "Hermitian matrix" means "symmetric matrix". Each A[i, j] = A[j, i]
- if matrix is complex then each A[i, j] must be a complex conjugate of A[j, i]

There are many situations when we need to know if matrix is real
And for that we need to know if the number is real.

So I think that we need this:
```Smalltalk
0.5 isRealNumber. "true"
3 isRealNumber. "true"
(2 - 7i) isRealNumber. "false"
```
Which is different from `isFloat`

SergeStinckwich (approved):

PR #196, opened 2021-02-24
kurtosis, skewness and variance are not defined on Cauchy distribution

SergeStinckwich:


nicolas-cellier-aka-nice:
While at it, a class comment would be welcome.
Note that standardDeviation answer nil while variance raise an Error.
Average is also questionnable. Strictly speaking, the integral is not defined...

SergeStinckwich:
> While at it, a class comment would be welcome.
> Note that standardDeviation answer nil while variance raise an Error.
> Average is also questionnable. Strictly speaking, the integral is not defined...

I add minimal comment for the class. You are welcome to add more :-)
I modify standardDeviation to be not defined also.

PR #208, opened 2021-06-10
check for invalid shapes in strassen multiplication

rakki-18:
Strassen multiplication between two matrices is only possible when the dimensions of both the matrices are a power of 2.
Currently, we get an error when we try to do `strassenProductWithMatrix:` with a matrix of dimension that is not a power of 2.

Changes introduced with this PR:
- An assertion condition is added to the `strassenProductWithMatrix:` method that checks that the shapes of the matrices are a power of 2.
- Unit test to check if the assertion failure is raised when `stassenProductWithMatrix` is called on invalid shapes.
- Unit test that checks if the output of Strassen multiplication on two valid matrices is correct

A better solution for this [problem](https://cs.stackexchange.com/questions/97998/strassens-matrix-multiplication-algorithm-when-n-is-not-a-power-of-2) would be to pad the matrix with zeros and make the dimensions to be a power of two.

### DIFF
--- a/src/Math-Matrix/PMMatrix.class.st
+++ b/src/Math-Matrix/PMMatrix.class.st
@@ -950,8 +950,13 @@ PMMatrix >> squared [
 PMMatrix >> strassenProductWithMatrix: aMatrix [
 	"Private"
 	| matrixSplit selfSplit p1 p2 p3 p4 p5 p6 p7 |
+
 	( self numberOfRows > 2 and: [ self numberOfColumns > 2])
 		ifFalse:[ ^self class rows: ( aMatrix rowsCollect: [ :row | self columnsCollect: [ :col | row * col]])].
+	
+	self assert: [ self numberOfRows isPowerOfTwo and: self numberOfColumns isPowerOfTwo  ] description: 'Matrix size should be a power of two'.
+	self assert: [ aMatrix numberOfRows isPowerOfTwo and: aMatrix numberOfColumns isPowerOfTwo  ] description: 'Matrix size should be a power of two'.
+		
 	selfSplit := self split.
 	matrixSplit := aMatrix split.
 	p1 := ( ( selfSplit at: 2) - ( selfSplit at: 4)) strassenProductWithMatrix: ( matrixSplit at: 1).
--- a/src/Math-Tests-Matrix/PMMatrixTest.class.st
+++ b/src/Math-Tests-Matrix/PMMatrixTest.class.st
@@ -742,7 +742,7 @@ PMMatrixTest >> testStrassenProductWithMatrix [
 		
 	| aPMMatrix expected strassenProduct |
 	
-   aPMMatrix := PMMatrix rows: #( (1 2 3 4) (1 0 1 2) (2 3 1 4) (4 3 2 2)).
+	aPMMatrix := PMMatrix rows: #( (1 2 3 4) (1 0 1 2) (2 3 1 4) (4 3 2 2)).
 	expected  := PMMatrix rows: #( (30 12 27 24) (12 6 11 10) (27 11 30 27) (24 10 27 33)).
 	strassenProduct := aPMMatrix transpose strassenProductWithMatrix: aPMMatrix.
 	self assert: strassenProduct equals: expected.
@@ -754,7 +754,7 @@ PMMatrixTest >> testStrassenProductWithMatrixWithInvalidShapes [
 	| aPMMatrix |
 	
 	"all the dimension of the matrices should be a power of 2 for strassen multiplication"
-   aPMMatrix := PMMatrix rows: #( (1 2 3) (1 0 1) (2 3 1) (2 0 1)).
+	aPMMatrix := PMMatrix rows: #( (1 2 3) (1 0 1) (2 3 1) (2 0 1)).
 	
 	self should: [ aPMMatrix strassenProductWithMatrix: aPMMatrix transpose ] raise: AssertionFailure.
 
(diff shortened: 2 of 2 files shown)


src/Math-Matrix/PMMatrix.class.st
```diff
@@ -950,8 +950,13 @@ PMMatrix >> squared [
 PMMatrix >> strassenProductWithMatrix: aMatrix [
 	"Private"
 	| matrixSplit selfSplit p1 p2 p3 p4 p5 p6 p7 |
+
 	( self numberOfRows > 2 and: [ self numberOfColumns > 2])
 		ifFalse:[ ^self class rows: ( aMatrix rowsCollect: [ :row | self columnsCollect: [ :col | row * col]])].
+	
+	self assert: [ self numberOfRows isPowerOfTwo and: self numberOfColumns isPowerOfTwo  ] description: 'Matrix size should be a power of two'.
+	self assert: [ aMatrix numberOfRows isPowerOfTwo and: aMatrix numberOfColumns isPowerOfTwo  ] description: 'Matrix size should be a power of two'.
```
hemalvarambhia, line 958: I like the use of assertions here. Pragmatic Programmers talks about using this to check pre-conditions in production

src/Math-Tests-Matrix/PMMatrixTest.class.st
```diff
@@ -737,6 +737,29 @@ PMMatrixTest >> testSkalarMultiplication [
...
 
+{ #category : #tests }
+PMMatrixTest >> testStrassenProductWithMatrix [
+		
+	| aPMMatrix expected strassenProduct |
+	
+   aPMMatrix := PMMatrix rows: #( (1 2 3 4) (1 0 1 2) (2 3 1 4) (4 3 2 2)).
+	expected  := PMMatrix rows: #( (30 12 27 24) (12 6 11 10) (27 11 30 27) (24 10 27 33)).
+	strassenProduct := aPMMatrix transpose strassenProductWithMatrix: aPMMatrix.
+	self assert: expected  equals: strassenProduct.
```
hemalvarambhia, line 748: This, I think, has to be the other way around. `self assert: strassenProduct equals: expected`, which reads better.

I really like the variable naming. Thank you

hemalvarambhia (approved):
Really nice PR.

hemalvarambhia:
@rakki-18 
> A better solution for this problem would be to pad the matrix with zeros and make the dimensions to be a power of two.

Will that be your next PR?

src/Math-Tests-Matrix/PMMatrixTest.class.st
rakki-18: @hemalvarambhia I have changed the order of `strassenProduct` and `expected`. Thank you.

rakki-18:
> @rakki-18
> 
> > A better solution for this problem would be to pad the matrix with zeros and make the dimensions to be a power of two.
> 
> Will that be your next PR?

I could try this.

src/Math-Tests-Matrix/PMMatrixTest.class.st
```diff
@@ -737,6 +737,29 @@ PMMatrixTest >> testSkalarMultiplication [
 	self assert: r class equals: PMSymmetricMatrix
 ]
 
+{ #category : #tests }
+PMMatrixTest >> testStrassenProductWithMatrix [
+		
+	| aPMMatrix expected strassenProduct |
+	
+   aPMMatrix := PMMatrix rows: #( (1 2 3 4) (1 0 1 2) (2 3 1 4) (4 3 2 2)).
+	expected  := PMMatrix rows: #( (30 12 27 24) (12 6 11 10) (27 11 30 27) (24 10 27 33)).
```
hemalvarambhia, line 746: Is this aligned with L345. In github it looks like it's been indented, @rakki-18

rakki-18: @hemalvarambhia I have fixed the indentation now.

hemalvarambhia:
@rakki-18 merged. Experiment with padding the matrix with zeroes. In your code, if there is reference on why this is done, add it as a comment. It'll help get a domain understanding. 

Thank you for hard work on this.

src/Math-Tests-Matrix/PMMatrixTest.class.st
```diff
@@ -737,6 +737,29 @@ PMMatrixTest >> testSkalarMultiplication [
...
+
+{ #category : #tests }
+PMMatrixTest >> testStrassenProductWithMatrixWithInvalidShapes [
+		
+	| aPMMatrix |
+	
+	"all the dimension of the matrices should be a power of 2 for strassen multiplication"
+	aPMMatrix := PMMatrix rows: #( (1 2 3) (1 0 1) (2 3 1) (2 0 1)).
+	
+	self should: [ aPMMatrix strassenProductWithMatrix: aPMMatrix transpose ] raise: AssertionFailure.
```
hemalvarambhia, line 759: @rakki-18 I just thought of something that maybe worth experimenting with. With the `assert: description:` message, is there a way to have it raise a more specific exception, such as `MatrixHasInvalidShape` - this helps with debugging, but has the disadvantage of YAGNI, unless the kind of shape checking done here is also done elsewhere, meaning a possible refactoring opportunity.

rakki-18: @hemalvarambhia We could try making a `MatrixHasInvalidShape` exception class. But as you mentioned, I could only find "matrix multiplication" where this class could be potentially used.

hemalvarambhia: Explore the idea, if you can, then, to examine the consequences for the design and debugging

rakki-18: Yeah, I will try to explore.

PR #210, opened 2021-07-17
Addition of class and operations on tensors

YvanGuifo:


SergeStinckwich:
Thank you Yvan, the Rank2 class was mostly a test. Can you commit instead class Tensor?

SergeStinckwich:
PMTensorRank2 was a test, can you commit only PMTensorRank and tests?

YvanGuifo:
In this pull request I integrated the classes
- PMTensor
- PMTensorTest
And the class for handling errors

SergeStinckwich:
Can you remove PMTensorRank2 and PMTensorRank2Test in your PR?

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,145 @@
...
+		'array',
+		'shape',
+		'first',
+		'strides'
+	],
+	#category : #'Math-Matrix'
+}
+
+{ #category : #'instance creation' }
+PMTensor class >> newWith: anInteger [
```
hemalvarambhia, line 14: I would appreciate this being a little more intention revealing: is this the dimension of the tensor? The rank?

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,145 @@
...
+{ #category : #public }
+PMTensor >> get: coords [
+
+	| position |
+	position := self getPosition: coords.
+	^ array at: position
+]
+
+{ #category : #'primitives - file' }
+PMTensor >> getPosition: coords [
```
hemalvarambhia, line 70: From the test, this could be elementAt

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,145 @@
...
+
+	^ shape
+]
+
+{ #category : #accessing }
+PMTensor >> size [
+
+	| product |
+	product := 1.
+	shape do: [ :each | product := each * product ].
```
hemalvarambhia, line 108: You could use `inject` here

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,145 @@
...
+}
+
+{ #category : #'instance creation' }
+PMTensor class >> newWith: anInteger [
+
+	^ self new array: {anInteger} withShape: #(  )
+]
+
+{ #category : #'instance creation' }
+PMTensor class >> rows: anArray [
```
hemalvarambhia, line 20: It would help me understand this message better if the argument name was clearer

YvanGuifo: This is due to the fact that we choose to browse the tensors according to the rows

hemalvarambhia: So what does `anArray` mean? Are you really referring to a row? If that is true, a rename would support my understanding

src/Math-Matrix/PMTensor.class.st
SergeStinckwich: This is a method to construct rank 0 (scalar) vector.

src/Math-Matrix/PMTensor.class.st
khinsen: I agree this name is misleading. Also,for ranks > 2, there are no rows or columns any  more.

Suggestion: fromNestedArray

src/Math-Matrix/PMTensor.class.st
khinsen: The tests are misleading. The return value is the index into the  flattened  version.

suggestion: flattenedIndexOf:

SergeStinckwich (changes requested):

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,154 @@
...
+	shape ifNotEmpty: [ self updateStrides]
+]
+
+{ #category : #accessing }
+PMTensor >> first [ 
+^first
+]
+
+{ #category : #'as yet unclassified' }
+PMTensor >> flattenedIndexOf: aFlatArray withShape: aShape [
```
SergeStinckwich, line 62: no, this idea is to replace getPosition: by  flattenedIndexOf:

SergeStinckwich (changes requested):

src/Math-Matrix/PMTensor.class.st
YvanGuifo: It is done

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+}
+
+{ #category : #'instance creation' }
+PMTensor class >> fromNestedArray: anArray [
+
+	^ self new array: anArray flattened withShape: (self shape:anArray)
+]
+
+{ #category : #'instance creation' }
+PMTensor class >> newWith: anInteger [
```
khinsen, line 20: How about `fromScalar:` for this method, in analogy to the newly named `fromNestedArray:`?

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+
+{ #category : #comparing }
+PMTensor >> = aTensor [
+
+	^ array = aTensor array & (first = aTensor first)
+	  & (strides = aTensor strides) & (shape = aTensor shape)
+]
+
+{ #category : #accessing }
+PMTensor >> array [
```
khinsen, line 43: This method looks like an accessor, but I don't think it should be one. The use of an array to store the elements is an implementation detail. What users might want is a conversion to a (flat) array, which I'd then call `asArray`. I'd then rename the instance variable `array` to `elements` and have an accessor of that name for internal use (marked "private").

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+PMTensor >> array: aFlatArray withShape: aShape [
+
+	array := aFlatArray.
+	shape := aShape copy.
+	self updateFirst.
+	shape ifNotEmpty: [ self updateStrides]
+]
+
+{ #category : #accessing }
+PMTensor >> first [
```
khinsen, line 57: I'd expect a method called `first` to be similar to `Collection >> first`, i.e. be equivalent to `at: 1`. Maybe the instance variable `first` should be renamed as well to get a better name for its accessor.

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+PMTensor >> fromNestedArray: aFlatArray withShape: aShape [
+
+	array := aFlatArray.
+	shape := aShape copy.
+	self updateFirst.
+	shape ifNotEmpty: [ self updateStrides]
+]
+
+{ #category : #public }
+PMTensor >> get: coords [
```
khinsen, line 81: Is there a good reason for not calling this `at:`?

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+
+{ #category : #'as yet unclassified' }
+PMTensor >> reshape: aNewShape [
+
+	^ self viewWithShape: aNewShape.
+
+]
+
+{ #category : #initialization }
+PMTensor >> set: coords value: aValue [
```
khinsen, line 101: Is there a good reason for not calling this `at:put:`?

src/Math-Matrix/PMTensorTest.class.st
```diff
@@ -0,0 +1,191 @@
...
+
+	| a b |
+	a := PMTensor fromNestedArray: (1 to: 6) asArray.
+	self assert: a first equals: #( 1).
+	b := a reshape: #( 3 2 ).
+	self assert: b first equals: #( 1 1 )
+]
+
+{ #category : #tests }
+PMTensorTest >> testFlattenedIndexOf [
```
khinsen, line 46: The example values used in this test case are not well chosen, because they are `1 to: n` reshaped. This means that `flattenedIndexOf:` and `get:` return the same values, making is harder to understand what is going on.

khinsen (changes requested):
This looks very good overall, except for same name choices that could be improved.

src/Math-Matrix/PMTensor.class.st
SergeStinckwich: first came from Numpy's naming but I agree this is colliding with Collection>>first Idea for a better name?

src/Math-Matrix/PMTensor.class.st
SergeStinckwich: I agree

src/Math-Matrix/PMTensor.class.st
SergeStinckwich: Agree also

src/Math-Matrix/PMTensor.class.st
```diff
@@ -0,0 +1,152 @@
...
+	^ array at: position
+]
+
+{ #category : #accessing }
+PMTensor >> rank [
+	^ shape size
+]
+
+{ #category : #'as yet unclassified' }
+PMTensor >> reshape: aNewShape [
```
hemalvarambhia, line 94: Normally you can contract tensors (check the literature) such that rank 2 tensors become rank 0 eg trace. Is this the intent here?

SergeStinckwich: no this this is reshape like in Numpy: https://numpy.org/doc/stable/reference/generated/numpy.reshape.html
Changing the dimension of the n-dimensional arrays without changing the contents.

hemalvarambhia: Ah, so we're copying numpy completely. Non blocking, but not sure if that's wise choice. Is the code quality is good? If it is, fine, go ahead. If not, then do we want those practices brought over?

SergeStinckwich: We are not copying completely. This is more an inspiration especially for the internal implementation as one 1D array

hemalvarambhia: At this point I would advise we choose wisely.

khinsen: Bias warning: I was part of the team that designed the basic NumPy Python API (back when it was called `Numeric`).

The basics of NumPy were heavily inspired by APL, and even 25 years later I still believe that was a good choice, and so I think we should follow the same path here. Over the years, the NumPy API got bloated, so I definitely don't recommend to follow NumPy everywhere.

Your reference to contraction points to the real problem: it's not a good idea to use the term "tensor" for what is really an N-dimensional array without any implied semantics.

hemalvarambhia: I am inclined to agree. Tensors are have a defined algebra and if this isn't a Tensor, then let's correct that. I agree that we should use Numpy terminology everywhere (where the naming is clear and domain-driven, we should).

Generally, where we can let's choose the names from Mathematics where that is possible.

hemalvarambhia: @khinsen What is APL?

khinsen: The very first interactive language based on N-dimensional arrays. Developed by Ken Iverson at IBM, first published in the 1960s. Details: https://en.wikipedia.org/wiki/APL_(programming_language)

YvanGuifo: Thanks. @hemalvarambhia what would be the appropriate name for this class ?

hemalvarambhia: How about MultidimensionalArray ? I get the feeling this may be what @khinsen meant

YvanGuifo: Thank you. I think I need to discuss it with @khinsen and @SergeStinckwich to make the choice.

SergeStinckwich: ``PMArray`` is not informative enough I guess, ``PMNDArray`` (n-dimensional array) is better.

SergeStinckwich:
@YvanGuifo there is still some changes (6) requested. Can you fix them, so I can accept the PR?

SergeStinckwich (approved):

PR #213, opened 2021-08-02
Addition of matrix product

YvanGuifo:
It contains :
- A method allowing to check that the matrices which one wishes to multiply have the same shape.
- A method to calculate the Hadamard product
- And test methods for this Hadamard product

### DIFF
--- a/src/Math-Matrix/PMNDArray.class.st
+++ b/src/Math-Matrix/PMNDArray.class.st
@@ -39,10 +39,10 @@ ifTrue:[
 ]
 
 { #category : #comparing }
-PMNDArray >> = anArray [
+PMNDArray >> = aPMNDArray [
 
-	^ array = anArray asArray & (first = anArray first)
-	  & (strides = anArray strides) & (shape = anArray shape)
+	^ array = aPMNDArray asArray & (first = aPMNDArray first)
+	  & (strides = aPMNDArray strides) & (shape = aPMNDArray shape)
 ]
 
 { #category : #'as yet unclassified' }
--- a/src/Math-Matrix/PMNDArray.class.st
+++ b/src/Math-Matrix/PMNDArray.class.st
@@ -99,12 +99,9 @@ PMNDArray >> fromNestedArray: aFlatArray withShape: aShape [
 
 { #category : #operation }
 PMNDArray >> hadamardProduct: aPMNDArray [
-	(self hasSameShapeAs: aPMNDArray)
-	ifTrue:[
-		^ self class new array: ((self asArray) with: (aPMNDArray asArray) collect:[:a :b| a*b])
-		withShape: self shape.
-		]
-	ifFalse:[ShapeMismatch signal ]
+	^ self with: aPMNDArray collect: [:a :b| a*b]
+	
+	
 ]
 
 { #category : #testing }
@@ -170,4 +167,14 @@ PMNDArray >> viewWithShape: aNewShape [
 
 	^ PMNDArray new array: self asArray withShape: aNewShape
 
+]
+
+{ #category : #operation }
+PMNDArray >> with: aPMNDArray collect: aBlock [
+	(self hasSameShapeAs: aPMNDArray)
+		ifFalse:[ShapeMismatch signal ].
+	^ self class new array: ((self asArray) with: (aPMNDArray asArray) collect: aBlock)
+		withShape: self shape.
+	
+	
 ]
--- a/src/Math-Matrix/PMNDArray.class.st
+++ b/src/Math-Matrix/PMNDArray.class.st
@@ -41,7 +41,7 @@ ifTrue:[
 { #category : #comparing }
 PMNDArray >> = anArray [
 
-	^ array = anArray array & (first = anArray first)
+	^ array = anArray asArray & (first = anArray first)
 	  & (strides = anArray strides) & (shape = anArray shape)
 ]
 
@@ -95,6 +95,21 @@ PMNDArray >> fromNestedArray: aFlatArray withShape: aShape [
 	shape := aShape copy.
 	self updateFirst.
 	shape ifNotEmpty: [ self updateStrides]
+]
+
+{ #category : #operation }
+PMNDArray >> hadamardProduct: aPMNDArray [
+	(self hasSameShapeAs: aPMNDArray)
+	ifTrue:[
+		^ self class new array: ((self asArray) with: (aPMNDArray asArray) collect:[:a :b| a*b])
+		withShape: self shape.
+		]
+	ifFalse:[ShapeMismatch signal ]
+]
+
+{ #category : #testing }
+PMNDArray >> hasSameShapeAs: aPMNDArray [
+	^ self shape = aPMNDArray shape
 ]
 
 { #category : #accessing }
--- a/src/Math-Matrix/PMNDArrayTest.class.st
+++ b/src/Math-Matrix/PMNDArrayTest.class.st
@@ -140,7 +140,7 @@ self assert: (t1 hadamardProduct: t2) equals: (PMNDArray fromNestedArray:
 ]
 
 { #category : #tests }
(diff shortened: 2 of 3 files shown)


src/Math-Matrix/PMNDArray.class.st
```diff
@@ -41,7 +41,7 @@ ifTrue:[
 { #category : #comparing }
 PMNDArray >> = anArray [
```
SergeStinckwich, line 43: Maybe use ``aPMNDArray`` instead of ``anArray``, because this is not an instance of ``Array`` class.

SergeStinckwich (changes requested):

src/Math-Matrix/PMNDArray.class.st
YvanGuifo: Ok. Thanks

khinsen (commented):
This looks OK as an implementation of the Hadamard product.But is that what we want? I'd expect to have element-wise application for any binary method (or two-argument code block), not just multiplication. So instead of
```
   array1 hadamardProduct: array2
```
I'd expect to write
```
   array1 with: array2 collect: #'*'
```

YvanGuifo:
Thank you @khinsen.  Need to clarify. When you talk about binary operation. Why look for other operators when it is a product that we perform.

khinsen:
It's a matter of interface design. Years of practice with APL, NumPy, and derived languages have shown that elementwise application of binary operations is very frequent. Products are nothing special, they are not used more often than addition and substraction. So, rather than introducing a specific operation for the product, I'd introduce a general one for any two-argument function.

APL and NumPy go even further and use the plain binary operators for elementwise combination. The Hadamard product of `a` and `b` is simply `a * b` in NumPy (or `a × b` in APL).

src/Math-Matrix/PMNDArray.class.st
```diff
@@ -97,6 +97,21 @@ PMNDArray >> fromNestedArray: aFlatArray withShape: aShape [
 	shape ifNotEmpty: [ self updateStrides]
 ]
 
+{ #category : #operation }
+PMNDArray >> hadamardProduct: aPMNDArray [
+	(self hasSameShapeAs: aPMNDArray)
+	ifTrue:[
```
hemalvarambhia, line 103: This is the normal flow, is that correct? If so, consider moving this out of the block and have run after the check. `ifFalse` becomes a guard clause. The code, in my opinion, then becomes a little more maintainable.

src/Math-Matrix/PMNDArray.class.st
```diff
@@ -97,6 +97,21 @@ PMNDArray >> fromNestedArray: aFlatArray withShape: aShape [
...
+		^ self class new array: ((self asArray) with: (aPMNDArray asArray) collect:[:a :b| a*b])
+		withShape: self shape.
+		]
+	ifFalse:[ShapeMismatch signal ]
+]
+
+{ #category : #testing }
+PMNDArray >> hasSameShapeAs: aPMNDArray [
```
hemalvarambhia, line 111: Really nice name!

src/Math-Matrix/PMNDArrayTest.class.st
```diff
@@ -100,6 +100,53 @@ PMNDArrayTest >> testFromNestedArray [
...
+					).
+					
+				
+			
+		
+	
+]
+
+{ #category : #tests }
+PMNDArrayTest >> testHadamardProductWithDifferentShapes [
```
hemalvarambhia, line 143: In the name, my suggestion is to emphasise in the test method name, the fact that it fails.

SergeStinckwich:
> It's a matter of interface design. Years of practice with APL, NumPy, and derived languages have shown that elementwise application of binary operations is very frequent. Products are nothing special, they are not used more often than addition and substraction. So, rather than introducing a specific operation for the product, I'd introduce a general one for any two-argument function.
> 
> APL and NumPy go even further and use the plain binary operators for elementwise combination. The Hadamard product of `a` and `b` is simply `a * b` in NumPy (or `a × b` in APL).

I think the suggestion of Konrad, is to introduce the elementwise method: ``with:collect:`` and then refactor ``hadamardProduct`` accordingly.

src/Math-Matrix/PMNDArray.class.st
SergeStinckwich: Agree, something like:
```
(self hasSameShapeAs:aPMNDArray)
ifFalse:[ShapeMismatch signal].
^ self class new array: ((self asArray) with: (aPMNDArray asArray) collect:[:a :b| a*b])
		withShape: self shape.
```

```

File: src/Math-Matrix/PMNDArrayTest.class.st (hunk 4 above)
Comment (YvanGuifo) [reply src/Math-Matrix/PMNDArrayTest.class.st:143]: Ok. Thanks

**Review by hemalvarambhia — APPROVED** [review]
All good.

File: src/Math-Matrix/PMNDArray.class.st [hunk 5]
```diff
@@ -97,6 +97,18 @@ PMNDArray >> fromNestedArray: aFlatArray withShape: aShape [
 	shape ifNotEmpty: [ self updateStrides]
 ]
 
+{ #category : #operation }
+PMNDArray >> hadamardProduct: aPMNDArray [
+	^ self with: aPMNDArray collect: [:a :b| a*b]
```
hemalvarambhia, line 102: Spacing would help me in `a*b`

src/Math-Matrix/PMNDArrayTest.class.st
```diff
@@ -100,6 +100,53 @@ PMNDArrayTest >> testFromNestedArray [
 	
 ]
 
+{ #category : #tests }
+PMNDArrayTest >> testHadamardProduct [
+|t1 t2|
+ t1 := PMNDArray fromNestedArray:  #(
```
hemalvarambhia, line 106: Suggestion: In the tests, see if you can give these better names

src/Math-Matrix/PMNDArrayTest.class.st
```diff
@@ -100,6 +100,53 @@ PMNDArrayTest >> testFromNestedArray [
...
+					#( #( 9 10) #( 11 12))
+					).
+					
+ t2 := PMNDArray fromNestedArray:
+				 #( 
+					#( #( 1 3 ) #( 2 1)) 
+					#( #( 0 6 ) #( 3 8))
+					#( #( 3 5) #( 1 10))
+					).
+self assert: (t1 hadamardProduct: t2) equals: (PMNDArray fromNestedArray:
```
hemalvarambhia, line 129: Suggestion: extract the expected outcome to an intention revealing variable e.g. `expectedHadamardProduct`

khinsen:
> I think the suggestion of Konrad, is to introduce the elementwise method: `with:collect:` and then refactor `hadamardProduct` accordingly.

Yes, exactly. Although I also wonder if we need the method `hadamardProduct` at all.

src/Math-Matrix/PMNDArrayTest.class.st
YvanGuifo: Hi @hemalvarambhia  you suggest I change `testHadamardProduct` to `expectedHadamardProduct`

hemalvarambhia: No, and I apologise for lack of clarity on my part. The `PMNDArray fromNestedArray...` part of the assertion should be extracted to a local variable called `expectedHadamardProduct`

YvanGuifo: Thank you @hemalvarambhia but I do not really understand the interest of defining a local variable in this test

YvanGuifo: OK. No, I understand the concerns of readability and documentation mentioned

SergeStinckwich (approved):

PR #214, opened 2021-08-02
Clean variables in SVD

SergeStinckwich:
Remove instance variables not used and also transform iv as temporary variables in SVD

### DIFF
--- a/src/Math-Matrix/PMSingularValueDecomposition.class.st
+++ b/src/Math-Matrix/PMSingularValueDecomposition.class.st
@@ -32,15 +32,11 @@ Class {
 	#name : #PMSingularValueDecomposition,
 	#superclass : #Object,
 	#instVars : [
-		'm',
-		'n',
 		'u',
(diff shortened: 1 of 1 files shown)


hemalvarambhia (approved):

src/Math-Matrix/PMSingularValueDecomposition.class.st
```diff
@@ -50,7 +46,7 @@ PMSingularValueDecomposition class >> decompose: aMatrix [
 
 { #category : #initialization }
 PMSingularValueDecomposition >> initialize: aMatrix [
-	| symU symV eigenU eigenV diag |
+	| symU symV eigenU eigenV diag m n |
```
hemalvarambhia, line 49: Maybe move these so that the order follows the actual assignments

SergeStinckwich: ok done

PR #224, opened 2021-10-07
fix #207 in cholesky decomposition

### LINKED ISSUE #207: cholesky decomposition fails for certain cases
I was trying to benchmark the Cholesky decomposition method and noticed that the method gives error for certain matrices.
Script to reproduce the error:
```Smalltalk
	|aPMMatrix expected|
	aPMMatrix := PMMatrix rows: #( #(4 20 8 2 18) #(20 116 80 18 126) #(8 80 117 32 135) #(2 18 32 73 109) #(18 126 135 109 272)).
        expected := PMMatrix rows: #( #(2 10 4 1 9) #(0 4 10 2 9) #(0 0 1 8 9) #(0 0 0 2 5) #(0 0 0 0 2)).
 
	aPMMatrix = (expected transpose * expected). "true"
        aPMMatrix choleskyDecomposition 
```
Calculating Cholesky decomposition for the above `aPMMatrix` gives the error, `sqrt undefined for number less than zero`.

```Smalltalk
diagonalValue := ((self at: j at: j) - rowSum) sqrt.
```
This line in the Cholesky decomposition method is causing the error where it is trying to calculate the square root of a negative number.

rakki-18:
## Issue
Fixes #207
A precedence error in the binary block of `#inject:into` was causing Cholesky decomposition to fail for matrices of size more than 3.
## Code before
```
[ sum+ (upperTriangular at: k at: i) * (upperTriangular at: k at: j) ]
```
 where `sum + (upperTriangular at: k at: i)` was being calculated first

## Why did tests pass before
In the tests, the matrix size was 3. 
So, the value of `sum` was always zero, hence the precedence did not matter.

## Change introduced
- Parenthesis has been added to take care of the precedence order
- New tests with a matrix size of more than 3 have been added

### DIFF
--- a/src/Math-Matrix/PMMatrix.class.st
+++ b/src/Math-Matrix/PMMatrix.class.st
@@ -366,15 +366,20 @@ PMMatrix >> choleskyDecomposition [
 
 					diagonalValue := ((self at: j at: j) - rowSum) sqrt.
 
+
 					upperTriangular at: j at: j put: diagonalValue ]
 				ifFalse: [
-					partialSum := (1 to: j - 1) inject: 0 into: [ :sum :k |
-						sum + (upperTriangular at: k at: i) * (upperTriangular at: k at: j) ].
 
+					partialSum := (1 to: j - 1) inject: 0 into: [ :sum :k |
+							sum + ((upperTriangular at: k at: i) * (upperTriangular at: k at: j)).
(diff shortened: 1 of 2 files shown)


rakki-18:
@olekscode Can you please review this when you are free?

Issue #226, opened 2021-12-10
take: method has been deprecated in Pharo


jordanmontt:
In Pharo 9 the method `take:` has the following:

```st
self
    deprecated: 'This method was renamed to #numberOfCombinationsTaken:'
    transformWith: '`@rec take: `@arg' -> '`@rec numberOfCombinationsTaken: `@arg'.
```

And in Pharo 10, the method `take:` has been completely removed.

So, if you try to use `PMMatrix>>take:` in a Pharo 10 image it fails with a DNU error. Because `PMMatrix>>take:` uses `SmallInteger>>take:` which has been removed.

jordanmontt:
Solved in #227

PR #227, opened 2021-12-10
Replaced take: calls with numberOfCombinationsTaken:

jordanmontt:
Solves issue #226


jordanmontt:
This changes will not work on Pharo 8 images.

PR #229, opened 2022-02-18
Refactorings in Math-ODE

SergeStinckwich:
- Refactor isNil ifTrue: as ifNil:
- Demos with announcement are working again


hemalvarambhia (commented):
I am still reviewing at the moment. I do notice duplication in the error handing parts and another thing. These will be worth addressing either now or in the next PR.

src/Math-ODE/PMAB2Stepper.class.st
```diff
@@ -15,9 +15,10 @@ PMAB2Stepper class >> order [
 
 { #category : #stepping }
 PMAB2Stepper >> doStep: aState prevState: prevState time: t [
-	self stepSize isNil
-		ifTrue: [ self error: 'step size required by stepper' ].
-	^ (self stepSize / 2) * (3 * (system state: aState time: t) - (system state: prevState time: t - self stepSize)) + aState
+
+	self stepSize ifNil: [ self error: 'step size required by stepper' ].
+	^ self stepSize / 2 * (3 * (system state: aState time: t)
```
hemalvarambhia, line 20: It looks like this computation may represent the same concept, but each class performs it differently. Is that right? At the very least, togive it a name,this computation might be extracted to an intention-revealing method or a local variable.

hemalvarambhia (approved):
A great start. Some suggestions offered on how to improve things next. Thank you 😊

src/Math-ODE/PMAB2Stepper.class.st
hemalvarambhia: Also, I think you might be able to refactor these methods towards the template design pattern, and then you might be led to this method being pulled up to the parent class, removing much more duplication, and subclasses providing behaviour to the parent class. I think this may simplify the code significantly. Thus I would encourage you to explore/introduce that in this PR or in the next.

src/Math-ODE/PMAB2Stepper.class.st
```diff
@@ -15,9 +15,10 @@ PMAB2Stepper class >> order [
 
 { #category : #stepping }
 PMAB2Stepper >> doStep: aState prevState: prevState time: t [
-	self stepSize isNil
-		ifTrue: [ self error: 'step size required by stepper' ].
-	^ (self stepSize / 2) * (3 * (system state: aState time: t) - (system state: prevState time: t - self stepSize)) + aState
+
```
hemalvarambhia, line 18: The doStep, could be renamed to

`stepFrom: previousState to: newState inStepsOf: stepSize` 

This seems more intention-revealing and almost fluent. Is this really what's happening? Are we stepping from one state to another in steps? Seems so.

src/Math-ODE/PMAB4Stepper.class.st
```diff
@@ -15,11 +15,18 @@ PMAB4Stepper class >> order [
 ]
 
 { #category : #stepping }
-PMAB4Stepper >> doStep3State: thirdState secondState:secondState firstState: firstState initState: initState initTime: initTime [
-
-	self stepSize isNil
-		ifTrue: [ self error: 'step size required by stepper' ].
-	 ^ self stepSize * (  ( (55/24) * (system state: thirdState time:  initTime+ (3* self stepSize) )) - ((59/24)*(system state: secondState time: initTime+ (2* self stepSize))) + ((37/24)*(system state: firstState time: initTime+ self stepSize ) )-   ((3/8)*(system state: initState time: initTime ) )) + thirdState.
+PMAB4Stepper >> doStep3State: thirdState secondState: secondState firstState: firstState initState: initState initTime: initTime [
```
hemalvarambhia, line 18: I wonder if we should make the signatures (selectors) look similar across the classes. There is a pattern by Sandi Metz, J B Rainsberger and Katrina Owen which involves making similar code look even more similar (i.e. makes the duplication obvious).

Issue #192, opened 2021-02-23
Rename PMComplex >> conjugated to conjugated 


olekscode:

hemalvarambhia:
@olekscode would we also want to rename the `conjugate` message in the `Quaternion` class?

hemalvarambhia:
<img width="698" alt="Screenshot 2022-04-10 at 21 50 50" src="https://user-images.githubusercontent.com/2665571/162639271-9de4380e-8c6d-4aad-a79c-4e0b81f400ed.png">
<img width="702" alt="Screenshot 2022-04-10 at 21 50 34" src="https://user-images.githubusercontent.com/2665571/162639272-a3c7587c-69be-4d2a-bbb9-00dc27a4914b.png">

hemalvarambhia:
I am noticing now that the Complex Number Test Class has a large number of methods. A refactor might split this up, and there are a number of ways to do that (I'll provide a list in a moment...).

nicolas-cellier-aka-nice:
Because quaternions also have a conjugate operation, will you need yet another `quaternionConjugate` selector?

hemalvarambhia:
I referred to Wikipedia, and the word used there is conjugation:

> Conjugation of quaternions is analogous to conjugation of complex numbers and to transposition (also known as reversal) of elements of Clifford algebras. To define it, let {\displaystyle q=a+b\,\mathbf {i} +c\,\mathbf {j} +d\,\mathbf {k} } be a quaternion. The conjugate of q is the quaternion {\displaystyle q^{*}=a-b\,\mathbf {i} -c\,\mathbf {j} -d\,\mathbf {k} }. It is denoted by q∗, qt, {\displaystyle {\tilde {q}}}, or q.[7] Conjugation is an involution, meaning that it is its own inverse, so conjugating an element twice returns the original element. The conjugate of a product of two quaternions is the product of the conjugates in the reverse order. That is, if p and q are quaternions, then (pq)∗ = q∗p∗, not p∗q∗.

The conjugation of a quaternion, in stark contrast to the complex setting, can be expressed with multiplication and addition of quaternions:

{\displaystyle q^{*}=-{\frac {1}{2}}(q+\,\mathbf {i} \,q\,\mathbf {i} +\,\mathbf {j} \,q\,\mathbf {j} +\,\mathbf {k} \,q\,\mathbf {k} )~.}

hemalvarambhia:
Are we discovering that conjugate may have meanings in different contexts?

PR #231, opened 2022-04-04
Refactor: Introduce complexConjugate Message, Deprecate conjugated

hemalvarambhia:
Issue no. #192.

### DIFF
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -179,7 +179,7 @@ PMComplex >> / anObject [
 		d := anObject imaginary.
 		newReal := ((a * c) + (b * d)) / ((c * c) + (d * d)).
 		newImaginary := ((b * c) - (a * d)) / ((c * c) + (d * d)).
-		^ PMComplex real: newReal imaginary: newImaginary].
+		^ self class real: newReal imaginary: newImaginary].
 	^ anObject adaptToComplex: self andSend: #/.
 ]
 
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -557,7 +557,7 @@ PMComplex >> isNumber [
 	^ true
 ]
 
-{ #category : #'mathematical functions' }
+{ #category : #testing }
 PMComplex >> isReal [
 
 	^ imaginary = 0 and: [ real >= 0 ]
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -404,10 +404,17 @@ PMComplex >> asComplex [
 ]
 
 { #category : #arithmetic }
+PMComplex >> complexConjugate [
+
+	^ self class real: real imaginary: imaginary negated
+]
+
+{ #category : #arithmetic }
 PMComplex >> conjugated [
+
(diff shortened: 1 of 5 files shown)


src/Math-Complex/PMComplex.class.st
```diff
@@ -544,6 +557,12 @@ PMComplex >> isNumber [
 	^ true
 ]
 
+{ #category : #testing }
+PMComplex >> isReal [
+
+	^ imaginary = 0 and: [ real >= 0 ]
+]
```
SergeStinckwich, line 564: 0 is not real?

src/Math-Complex/PMComplex.class.st
```diff
@@ -694,14 +713,14 @@ PMComplex >> sinh [
 
 { #category : #'mathematical functions' }
 PMComplex >> sqrt [
+
 	"Return the square root of the receiver with a positive imaginary part."
 
 	| u v |
-	(imaginary = 0 and: [real >= 0])
-		ifTrue:	[^self class real: real sqrt imaginary: 0].
+	self isReal ifTrue: [ ^ self class real: real sqrt imaginary: 0 ].
```
SergeStinckwich, line 720: I think you extract ``(imaginary = 0 and: [real >= 0])`` a bit too fast, this is not replaced by isReal but by isReal and positive.

hemalvarambhia: How about `z isZero or: z isReal` ?

hemalvarambhia: How about `z isZero or: z isReal`?

src/Math-Complex/PMComplex.class.st
hemalvarambhia: Very true. Thus we have a bug.

src/Math-Complex/PMComplex.class.st
SergeStinckwich: I don't know if this is necessary. Maybe keep it like it was before.

hemalvarambhia: Sure. Perhaps we need more tests before refactoring.

SergeStinckwich (approved):

src/Math-Complex/PMComplex.class.st
hemalvarambhia: I have inlined the method for us.

src/Math-Complex/PMComplex.class.st
```diff
@@ -694,14 +707,15 @@ PMComplex >> sinh [
...
 
 	| u v |
-	(imaginary = 0 and: [real >= 0])
-		ifTrue:	[^self class real: real sqrt imaginary: 0].
+	(imaginary = 0 and: [ real >= 0 ]) ifTrue: [ 
+		^ self class real: real sqrt imaginary: 0 ].
 	v := (self abs - real / 2) sqrt.
 	u := imaginary / 2 / v.
-	^self class real: u imaginary: v
+	^ self class real: u imaginary: v
```
hemalvarambhia, line 718: These may be formatting improvements by Pharo when I inlined the `isReal` method

hemalvarambhia:
@SergeStinckwich I think we could add more tests for the square root message. I can do that as my next PR. Rather than tests in one method, I would rather write separate test methods with good names.

PR #237, opened 2022-04-12
Fixes #236

### LINKED ISSUE #236: Rename instance variables of PMSingularVectorDecomposition

jordanmontt:
Renamed instance variables with more descriptive names.

### DIFF
--- a/src/Math-Matrix/PMSingularValueDecomposition.class.st
+++ b/src/Math-Matrix/PMSingularValueDecomposition.class.st
@@ -32,9 +32,9 @@ Class {
 	#name : #PMSingularValueDecomposition,
 	#superclass : #Object,
 	#instVars : [
-		'leftSingularForm',
-		'rightSingularForm',
-		'sForm'
+		'diagonalSingularValueMatrix',
+		'leftSingularMatrix',
+		'rightSingularMatrix'
 	],
 	#category : #'Math-Matrix'
 }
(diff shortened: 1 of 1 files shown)


src/Math-Matrix/PMSingularValueDecomposition.class.st
```diff
@@ -56,26 +56,26 @@ PMSingularValueDecomposition >> initialize: aMatrix [
...
-	v := (PMMatrix rows: eigenV vectors) transpose.
+	leftSingularForm := (PMMatrix rows: eigenU vectors) transpose.
+	rightSingularForm := (PMMatrix rows: eigenV vectors) transpose.
 	diag := m < n
 		        ifTrue: [ eigenU values ]
 		        ifFalse: [ eigenV values ].
-	s := PMMatrix rows: m columns: n random: 0.
-	s setDiagonal: diag sqrt
+	sForm := PMMatrix rows: m columns: n random: 0.
+	sForm setDiagonal: diag sqrt
```
hemalvarambhia, line 65: It looks like this is diagonalised square root. Is that right? If so, perhaps the name could be diagonalisedSquareRoot?

SergeStinckwich:
I understand that you want to be more explicit and give names, but I think the name U, S and V are used a lot in the literature.

jordanmontt:
Ok, it is just a suggestion. Feel free to close this PR if you think is not pertinent

olekscode:
Hello,
@SergeStinckwich is right, the names U, V, and S are used in all the literature.
However, I think that single-letter names are a very bad practice.
In Pharo, we always try to be verbose. And we never know who will be the person debugging and reading the code.
Chances are, this will be someone who is not familiar with the literature on the topic.

That is why, I propose to accept this pull request.

Besides, @jordanmontt has added a clear comments to the accessors, explaining that `rightSingularMatrix` is same as `V` and so on:
```st
PMSingularValueDecomposition >> rightSingularMatrix 
	"Matrix V in the A = USV decomposition. The columns of this matrix are the eigenvectors of the (A transposed) * A matrix (called right singular vectors). The columns of this matrix are orthonormal"

	^ rightSingularMatrix
```

olekscode (approved):

hemalvarambhia:
@SergeStinckwich @olekscode I'll merge this tomorrow 🌄

PR #238, opened 2022-04-13
Refactor Option 1: Break Up Complex Number Tests Into Binary and Unary Operation Based Test Cases

hemalvarambhia:
As the tests are a little big, I've begun breaking them up and, initially, I have broken them up into:
- binary operations ( x, +, %)
- unary operations (trigonometric functions, square root, etc.)

It does not feel ideal, so I will keep iterating.

#239 offers an lternative test design

Ref: [Structure and Interpretation of Test Cases, Kevlin Henney](https://youtu.be/tWn8RA_DEic)


hemalvarambhia:
I'm not a fan of option 1 anymore. I'd prefer to improve naming first (test messages and categorisation) and then look at break up strategies.

PR #239, opened 2022-04-13
Refactor: Group Complex Number Tests Into Well Named Categories.

hemalvarambhia:
In this PR, I groupes the test methods into well-named categories and added additional tests.

This is not the end- I will keep iterating. 

Ref: [Structure and Interpretation of Test Cases, Kevlin Henney](https://youtu.be/tWn8RA_DEic)


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -4,7 +4,7 @@ Class {
 	#category : #'Math-Tests-Complex'
 }
 
-{ #category : #tests }
+{ #category : #'unary operations' }
```
SergeStinckwich, line 7: I would prefer 'tests - unary operations'. Some library use this pattern.

SergeStinckwich (changes requested):

src/Math-Tests-Complex/PMComplexTest.class.st
hemalvarambhia: Cool. Is this something you'd like for the other categories ?

hemalvarambhia: I'll make the change in the next commit.

hemalvarambhia:
I have merge conflicts so I will close this and issue a new PR, taking @SergeStinckwich 's feedback into account

PR #242, opened 2022-04-19
Refactor testFFT

hemalvarambhia:
The test method is too long, and this is making it difficult to locate the precise part that is the reason. I've chosen to break it up in order to understand that better. 

Along the way, we have surfaced two tests that are similar, in that they take an input (cosine way or array), compute the discrete Fourier Transfer and check the size of the output signal. Perhaps we can remove the one that tests a large data set without loss of confidence.

### DIFF
--- a/src/Math-Tests-FastFourierTransform/PMFFTTest.class.st
+++ b/src/Math-Tests-FastFourierTransform/PMFFTTest.class.st
@@ -57,8 +57,8 @@ PMFFTTest >> testFastFourierTransformDoesNotLoseAnySignal [
 	f inverseTransform.
 	
 	zeroedImaginaryData := (f imaginaryData select: [ :i | i equalsTo: 0 ]) size.
-	self assert: zeroedImaginaryData equals: 256.
 	zeroedRealData := (f realData - cosineWaveSignal select: [ :i | i equalsTo: 0 ]) size.
+	self assert: zeroedImaginaryData equals: 256.
 	self assert: zeroedRealData equals: 256.
 	
 	PMMitchellMooreGenerator reset: 1.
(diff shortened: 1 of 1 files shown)


src/Math-Tests-FastFourierTransform/PMFFTTest.class.st
```diff
@@ -21,35 +28,64 @@ PMFFTTest >> testBitReverse [
...
-	data1 := data collect: [ :i | i + 0.001 random - 0.0005 ].
-	f := PMFastFourierTransform data: data1.
+
+	zeroedImaginaryData := (f imaginaryData select: [ :i | i equalsTo: 0 ]) size.
+	inputOutputDiff := f realData - inputSignal.
+	zeroedRealData := (inputOutputDiff select: [ :i | i equalsTo: 0 ]) size.
+	self assert: zeroedImaginaryData equals: 256.
+	self assert: zeroedRealData equals: 256.
+
+	PMMitchellMooreGenerator reset: 1
```
hemalvarambhia, line 46: @SergeStinckwich Was this copied and pasted in error?

src/Math-Tests-FastFourierTransform/PMFFTTest.class.st
```diff
@@ -21,35 +28,63 @@ PMFFTTest >> testBitReverse [
 ]
 
 { #category : #tests }
-PMFFTTest >> testFFT [
-	| data data1 f s |
-	data := (1 to: 256)
-		collect: [ :i | (Float pi * (i - 1) / (256 / 8)) cos ].
-	f := PMFastFourierTransform data: data.
+PMFFTTest >> testFastFourierTransformCalculatesTheDiscreteFourierTransformWithoutLossOfSignalForLargeDataSets [
```
olekscode, line 31: Wow, that's a long name :)

olekscode (approved):
I have low expertise here - thus low confidence

But I think those changes are good
The code is more clear than before and the tests are passing

PR #244, opened 2022-04-19
Tests failing on pharo 10

olekscode:
Fixed 9 tests that were failing on Pharo 10


hemalvarambhia (approved):
Lost some test coverage but how well would any tests protect us from regression?

olekscode:
I don't understand how there could be a loss of coverage... I didn't add any methods and didn't remove any tests.
I also didn't remove any calls to PolyMath methods

So it seems like a false alarm

hemalvarambhia:
Possibly. Merge at your leisure.

SergeStinckwich:
Thank you Oleks!

PR #245, opened 2022-04-19
Refactor: Separate Out Complex Conjugation Tests

hemalvarambhia:
Clarified test naming and grouped the tests into a intention-revealing category.

Added new tests that demonstrates the properties of complex conjugation e.g. involution, distributive.

### DIFF
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -546,7 +546,7 @@ PMComplex >> isComplexConjugateOfAComplexNumber: aComplexNumber [
 
 	"Answer true if self and aComplexNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
 
-	^ self complexConjugate = aComplexNumber
+	^ self = aComplexNumber complexConjugate
 ]
 
 { #category : #testing }
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -544,7 +544,9 @@ PMComplex >> isComplexConjugateOf: aNumber [
 { #category : #testing }
 PMComplex >> isComplexConjugateOfAComplexNumber: aComplexNumber [
 	"Answer true if self and aComplexNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
-	^ (aComplexNumber real = real) and: [ aComplexNumber imaginary = (-1 * imaginary) ]
+	| conjugate |
+	conjugate := self complexConjugate .
+	^ (aComplexNumber real = conjugate real) and: [ aComplexNumber imaginary = conjugate imaginary ]
 ]
 
 { #category : #testing }
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -543,10 +543,10 @@ PMComplex >> isComplexConjugateOf: aNumber [
 
 { #category : #testing }
 PMComplex >> isComplexConjugateOfAComplexNumber: aComplexNumber [
+
 	"Answer true if self and aComplexNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
-	| conjugate |
-	conjugate := self complexConjugate .
-	^ (aComplexNumber real = conjugate real) and: [ aComplexNumber imaginary = conjugate imaginary ]
+
+	^ self complexConjugate = aComplexNumber
 ]
 
 { #category : #testing }
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -780,7 +780,7 @@ PMComplexTest >> testThatComplexConjugationIsNotSymmetric [
 ]
 
 { #category : #'testing - complex conjugation' }
-PMComplexTest >> testThatComplexConjugationLeavesModulusUnchanged [
+PMComplexTest >> testThatComplexConjugationPreservesModulus [
 	| z |
 	z := 3 - 4 i.
 	
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -746,12 +746,6 @@ PMComplexTest >> testThatComplexConjugateOfAPureRealNumberIsItself [
 ]
 
 { #category : #'testing - complex conjugation' }
-PMComplexTest >> testThatComplexConjugateOfAnIntegerIsItself [
-
-	self assert: (5 isComplexConjugateOf: 5)
-]
-
-{ #category : #'testing - complex conjugation' }
 PMComplexTest >> testThatComplexConjugationIsAnInvolution [		
 	| z |
 	z := -9 + 24 i.
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -390,12 +390,6 @@ PMComplexTest >> testHash [
 ]
 
 { #category : #'testing - complex conjugation' }
-PMComplexTest >> testIsComplexConjugateOfDifferentReal [
-
-	self deny: (-5 isComplexConjugateOf: 5)
-]
-
-{ #category : #'testing - complex conjugation' }
 PMComplexTest >> testIsComplexConjugateOfNonConjugateComplexAndReal [
 
 	self deny: ((5 + 3i) isComplexConjugateOf: 5)
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -769,7 +769,7 @@ PMComplexTest >> testThatComplexConjugationIsDistributiveOverDivision [
 	z := 6 - 15 i.
 	w := 10 + 5 i.
 	
-	self assert: (z / w) complexConjugate equals: z complexConjugate / w complexConjugate
+	self assert: (z / w) complexConjugate equals: (z complexConjugate / w complexConjugate)
 ]
 
 { #category : #'testing - complex conjugation' }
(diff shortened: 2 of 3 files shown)


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +735,63 @@ PMComplexTest >> testTanh [
...
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatComplexConjugationPreservesModulus [
+	| z |
+	z := 3 - 4 i.
+	
+	self assert: z complexConjugate abs equals: 5
+]
+
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatComplexConjugationYieldsANumberWithTheSameRealPartAndNegativeImaginaryPart [
```
hemalvarambhia, line 791: Rename this

src/Math-Complex/PMComplex.class.st
```diff
@@ -543,8 +543,10 @@ PMComplex >> isComplexConjugateOf: aNumber [
 
 { #category : #testing }
 PMComplex >> isComplexConjugateOfAComplexNumber: aComplexNumber [
+
 	"Answer true if self and aComplexNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
-	^ (aComplexNumber real = real) and: [ aComplexNumber imaginary = (-1 * imaginary) ]
+
+	^ self complexConjugate = aComplexNumber
```
hemalvarambhia, line 549: What happens if we swap this around?

```
aComplexNumber complexConjugate = self
```

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +729,72 @@ PMComplexTest >> testTanh [
...
+PMComplexTest >> testThatComplexConjugationPreservesModulus [
+	| z |
+	z := 3 - 4 i.
+	
+	self assert: z complexConjugate abs equals: 5
+]
+
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatTwoEntirelyDifferentComplexNumbersAreNotComplexConjugatesOfEachOther [
+	| z |
```
hemalvarambhia, line 792: Not sure if this test is useful.

src/Math-Complex/PMComplex.class.st
```diff
@@ -543,8 +543,10 @@ PMComplex >> isComplexConjugateOf: aNumber [
 
 { #category : #testing }
 PMComplex >> isComplexConjugateOfAComplexNumber: aComplexNumber [
+
 	"Answer true if self and aComplexNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
-	^ (aComplexNumber real = real) and: [ aComplexNumber imaginary = (-1 * imaginary) ]
+
+	^ self = aComplexNumber complexConjugate
```
hemalvarambhia, line 549: Does this need to a method? It's only used by a Matrix to check it is Hermitian. I don't feel it's needed. Is there evidence to the contrary?

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +729,72 @@ PMComplexTest >> testTanh [
 	self assert: (c2 imaginary closeTo: c tanh imaginary).
 ]
 
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatComplexConjugateHasEqualRealPartAndAnImaginaryPartWithEqualMagnitudeButOppositeSign [
```
olekscode, line 733: Those long test names are very hard to read even if I can see all of it (and System Browser will hide most of the name because of the window size).
```
testThatComplexConjugateHasEqualRealPartAndAnImaginaryPartWithEqualMagnitudeButOppositeSign
```
I really prefer this name:
```
testComplexConjugate
```
Why?
Because we are testing if `complexConjugate` method returns the correct complex conjugate.
That's it.
Saying that _"it must have equal real part and an imaginary part with equal magnitude but opposite sign"_ is just the definition of a complex conjugate. This sentence should be in a documentation comment of the `complexConjugate` method but not in the test name

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +729,72 @@ PMComplexTest >> testTanh [
...
+	| z |
+	z := 3 + 2 i.
+	
+	self assert: z complexConjugate equals: 3 - 2 i
```
olekscode, line 737: I think that the code would be more readable if we don't create a variable here:
```st
self assert: (3 + 2i) complexConjugate equals: (3 - 2i).
```
Complex numbers are native to Pharo, they should be used like any other numbers.
And personally, I wouldn't write this:
```st
| a b |
a := 4.
b := -4.

self assert: a negated equals: b.
```
I would just write
```st
self assert: 4 negated equals: -4
```
So the same logic should apply to complex numbers

olekscode: To sum up, I propose the following:
```st
testComplexConjugate

    self assert: (3 + 2i) complexConjugate equals: (3 - 2i).
```

olekscode: The method with that name already exists but it is not good...
```st
testComplexConjugate
	| z complexConjugateOfZ expected |
	z := 5 - 6 i.
	
	complexConjugateOfZ := z complexConjugate .
	
	expected := 5 + 6 i.
	self assert: complexConjugateOfZ equals: expected.
```
I propose to replace it with a more readable one-line version:
```st
testComplexConjugate

	self assert: (5 - 6i) complexConjugate equals: (5 + 6i).
```

olekscode: I will open a separate issue for that

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +729,72 @@ PMComplexTest >> testTanh [
...
+]
+
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatComplexConjugateOfAPureRealNumberIsItself [
+	| z |
+	z := 5 + 0 i.
+	
+	self assert: z complexConjugate equals: 5.
```
olekscode, line 745: This test is not good.

First, the same comment about test name as above.
I would use a shorter name: `testComplexConjugateOfRealNumber`
The fact that it's itself is a definition of complex conjugate and should be in the method comment and not the test name

Second, about the temporary variable, same comment as above - I would not use it because it is unnecessary and make the code harder to read (of course, this is subjective).

Third (!), this test is misleading.
`(5 + 0i) complexConjugate` does not return `5` (a real number)
It returns `(5 + 0i)` (a complex number)

The reason why this test passes is because `(5 + 0i) = 5`

--------------
I would do the following:

1. Remove this test, because for complex numbers, testing that `self assert: (5 - 0i) complexConjugate equals (5 + 0i)` is the same test as `self assert: (3 + 2i) complexConjugate equals (3 - 2i)`. And we already have that test as `testComplexConjugate`
2. Add `complexConjugate` to `Number` class (see my first comment to this review)
3. Write the following test for `Number`:

```st
NumberTest >> testComplexConjugate

	self assert: 5 complexConjugate equals: 5
```

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -783,6 +729,72 @@ PMComplexTest >> testTanh [
...
+	
+	self assert: z complexConjugate abs equals: 5
+]
+
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testThatTwoEntirelyDifferentComplexNumbersAreNotComplexConjugatesOfEachOther [
+	| z |
+	z := -0.5 - 1 i.
+	
+	self deny: z complexConjugate equals: (3 - 2i)
```
olekscode, line 795: I would remove ALL those tests.
They pass **by definition** if `testComplexConjugate` passes

There is no need to pollute the code with 6 tests that can not possibly fail
The only way this can fail is either:

(a) if `testComplexConjugate` fails
(b) if complex numbers don't behave as expected, but this should be (an it is) reflected in the tests of complex numbers. It has nothing to do with complex conjugates specifically

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -810,6 +822,22 @@ PMComplexTest >> testTwoComplexNumbersWithDifferentRealPartsAreNotEqual [
  	self deny: z equals: w.
 ]
 
+{ #category : #'testing - complex conjugation' }
+PMComplexTest >> testWeCanIdentifyTwoComplexNumbersAsBeingComplexConjugatesOfEachOther [
+	| z |
+	z := 3 - 2 i.
+	
+	self assert: (z isComplexConjugateOf: (3 + 2 i))
```
olekscode, line 830: For the same reasons as discussed above, I would name this test `testIsComplexConjugateOf` and I would not create a temporary variable. So the code would be:

```st
testIsComplexConjugateOf

	self assert: ((3 - 2i) isComplexConjugateOf: (3 + 2i)).
```

But this was already in the tests that you propose to remove.

Also, look at those tests that you propose to remove... they cover many important scenario:

- is complex conjugate of complex and real
- is complex conjugate of real and complex (different receiver)
- is complex conjugate of complex and complex
- is complex conjugate of complex and complex when they are not complex conjugate of each other
...

If we keep the `isComplexConjugateOf:` method, we must keep all those tests.
But if we remove it, and add `complexConjugate` to `Number` as I proposed above (will be a separate issue), then we can remove them and simply add `testComplexConjugate` to both `Number` and `PMComplex`

olekscode (changes requested):
I propose to reject this PR. Please see my comments

src/Math-Tests-Complex/PMComplexTest.class.st
olekscode: Issue #246

src/Math-Complex/PMComplex.class.st
olekscode: Issue #247

src/Math-Tests-Complex/PMComplexTest.class.st
hemalvarambhia: There is a bit more to complex conjugation e.g. it is distributive over the binary operations for example hence:

I would say `testWeCanComputeTheComplexConjugate` 

but if you insist I'm happy to go with what you suggest.

hemalvarambhia: Also, there is the value or idea of **Tests as documentation** c.f. XUnit Test Patterns by Géerard Mezsaros

PR #249, opened 2022-04-23
Fixed #246. Refactored the PMComplexTest >> testComplexConjugate

### LINKED ISSUE #246: Refactor PMComplexTest >> testComplexConjugate
This is the current implementation
```st
testComplexConjugate
	| z complexConjugateOfZ expected |
	z := 5 - 6 i.
	
	complexConjugateOfZ := z complexConjugate .
	
	expected := 5 + 6 i.
	self assert: complexConjugateOfZ equals: expected.
```
It has many unnecessary temporary variables that make it hard to read
I propose to replace it with the following:
```st
testComplexConjugate

	self assert: (5 - 6i) complexConjugate equals: (5 + 6i).
```
Same thing but more clear

olekscode:


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -201,13 +201,8 @@ PMComplexTest >> testComplexCollection [
 
 { #category : #tests }
 PMComplexTest >> testComplexConjugate [
-	| z complexConjugateOfZ expected |
```
hemalvarambhia, line 204: Consider giving the category a better name, perhaps `arithmetic` (is complexConjugate in that category?).

hemalvarambhia (approved):

src/Math-Tests-Complex/PMComplexTest.class.st
olekscode: Yes, it's a good idea to have descriptive categories
I will merge this one and then we will fix all categories in PMComplex

PR #250, opened 2022-04-23
Fixed #248. Added a transformation rule to PMComplex >> conjugated

### LINKED ISSUE #248: Add a transformation rule to the deprecated method PMComplex >> conjugated
Now the deprecation looks like this:
```st
PMComplex >> conjugated
	"Return the complex conjugate of this complex number."

	self
		deprecated: 'Use #complexConjugate instead'
		on: '3 April 2022'
		in:
		'Pharo-9.0.0+build.1575.sha.9bb5f998e8a6d016ec7abde3ed09c4a60c0b4551 (64 Bit)'.

	^ self complexConjugate
```
I propose to add a transformation rule and to fix the version string (too much information there)
```st
PMComplex >> conjugated
	"Return the complex conjugate of this complex number."

	self
		deprecated: 'Use #complexConjugate instead'
		on: '3 April 2022'
		in: #Pharo9
		transformWith: '`@rec conjugated' -> '`@rec complexConjugate'.

	^ self complexConjugate
```

olekscode:


src/Math-Complex/PMComplex.extension.st
```diff
@@ -17,20 +17,20 @@ PMComplex >> productWithVector: aVector [
...
-
 { #category : #'*Math-Complex' }
 PMComplex >> random [
 	"analog to Number>>random. However, the only bound is that the abs of the produced complex is less than the length of the receive. The receiver effectively defines a disc within which the random element can be produced."
 	^ self class random * self
 	
 ]
 
+{ #category : #'*Math-Complex' }
+PMComplex class >> random [
```
hemalvarambhia, line 28: Is this used anywhere or has it been added to stay polymorphic to `Number`?

hemalvarambhia (approved):

src/Math-Complex/PMComplex.extension.st
olekscode: Ah, this one is strange
I didn't add it
And if you look at the diff, the same method is marked as added and removed
Nothing was changed

So I think it's a bug in how Github calculates the diff

PR #251, opened 2022-04-23
Fixed #247. Added complexConjugate to Number and removed isCompleConjugateOf:

### LINKED ISSUE #247: Add complexConjugate to Number and remove isComplexConjugateOf:
**Step 1:** We can add a `complexConjugate` method to the `Number` class that would return `self`
```st
Number >> complexConjugate
    "The complex conjugate of a complex number (a + bi) is another complex number (a - bi).
    Every real number x is also a complex number with imaginary part equal to 0.
    In other words, x = x + 0i and x = x - 0i.
    Therefore, the complex conjugate of a real number is the same real number"
    
    ^ self
```
**Step 2:** And a test:
```st
NumberTest >> testComplexConjugate

    self assert: 5 complexConjugate equals: 5
```
**Step 3:** Then inside `PMMatrix >> isHermitian`, instead of checking this condition
```st
(self at: i at: j) isComplexConjugateOf: (self at: j at: i)
```
We simply write this:
```st
(self at: i at: j) complexConjugate = (self at: j at: i)
```

**Step 4:** Then we deprecate the `isComplexConjugateOf:` because it is redundant:
```st
PMComplex >> isComplexConjugateOf: aNumber
    "Answer true if self and aNumber are complex conjugates of each other. The complex conjugate of a complex number is the number with an equal real part and an imaginary part equal in magnitude but opposite in sign."
    self
        deprecated: 'This method is redundant. Just check the equality with complexConjugate'
        transformWith: '`@rec isComplexConjugate: `@arg' -> '`@rec complexConjugate = `@arg'.
        
    ^ self complexConjugate = aNumber
```
And the same for `Number`
```st
Number >> isComplexConjugateOf: aNumber
    "A complex conjugate of a real number is same real number"
    self
        deprecated: 'This method is redundant. Just check the equality with complexConjugate'
        transformWith: '`@rec isComplexConjugate: `@arg' -> '`@rec complexConjugate = `@arg'.
        
    ^ self complexConjugate = aNumber
```
And we directly remove those two methods because they were only meant for double dispatch:

- `PMComplex >> isComplexConjugateOfAComplexNumber:`
- `Number >> isComplexConjugateOfAComplexNumber:`

**Step 5:** Finally, we remove the tests of `isComplexConjugateOf:` from `PMComplexTest` class. Those tests are not needed anymore and their job will be done by three tests: `PMComlexTest >> testComplexConjugate`, `NumberTest >> testComplexConjugate` and `PMComplexTest >> testEquality`.

Those are the tests to be removed:

- `testIsComplexConjugateOfConjugateComplex`
- `testIsComplexConjugateOfConjugateComplexAndReal`
- `testIsComplexConjugateOfConjugateRealAndComplex`
- `testIsComplexConjugateOfDifferentReal`
- `testIsComplexConjugateOfNonConjugateComplexAndReal`
- `testIsComplexConjugateOfNonConjugateDifferentComplex`
- `testIsComplexConjugateOfNonConjugateRealAndComplex`
- `testIsComplexConjugateOfSameComplex`
- `testIsComplexConjugateOfSameReal`

olekscode:
…ugateOf:


hemalvarambhia (approved):

hemalvarambhia:
This is better. Those tests you deleted didn't seem useful and were not well named.

The coverage has gone down. Do check whether the loss represents something significant before merging.

src/Math-Complex/Number.extension.st
```diff
@@ -15,7 +15,12 @@ Number >> asComplex [
 
 { #category : #'*Math-Complex' }
 Number >> complexConjugate [
-	^ self.
+	"The complex conjugate of a complex number (a + bi) is another complex number (a - bi).
```
hemalvarambhia, line 18: I think I already added this in a previous PR

hemalvarambhia (approved):
Other than the complexConjugate method being added again, this is great.

src/Math-Complex/Number.extension.st
hemalvarambhia, line 18: I wouldn't have added the comment here and instead used tests to document this (Tests as Documentation). I would only comment on a design decision I made, the why, if you will.

PR #252, opened 2022-04-23
Refactor FFT tests: Extract Duplicate Code To An Intention-Revealing Custom Assertion

hemalvarambhia:
Extracting duplicate to a method felt like the most conservative step forward and is an appeal to Kent Beck's Four Rules of Simple Design.

Naming the assertion is proving difficult and so it may be preferable to keep the duplication.


hemalvarambhia:
I chose to close this as I would prefer duplication after discovering I had trouble naming the custom assertion.

PR #253, opened 2022-04-23
Refactor complex number tests: Categorised the test and added other interesting test cases

hemalvarambhia:
- Added a category which emphasises that complex numbers have a polar coordinate form; 

- added missing tests that felt interesting enough to confirm:
--argument of a real number is 0 radians
-- argument of a negative pure imaginary number is - Pi /2 (I was expecting 3 Pi / 2)

### DIFF
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -163,12 +163,7 @@ PMComplexTest >> testArcTanDenominator [
 ]
 
 { #category : #'testing - polar coordinates' }
-PMComplexTest >> testArgumentOfAPositiveRealNumber [
-	self assert: (5 + 0 i) arg equals: 0.
-]
-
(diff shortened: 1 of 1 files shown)


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -162,8 +162,8 @@ PMComplexTest >> testArcTanDenominator [
 	self assert: (c2 arcTan: c1 * c1) equals: Float pi
 ]
 
-{ #category : #tests }
-PMComplexTest >> testArg [
+{ #category : #'testing - polar coordinates' }
+PMComplexTest >> testArgumentOfAPositivePureImaginaryNumber [
```
hemalvarambhia, line 166: Another test is cos pi/4 + i sin pi/4, whose argument is pi/4. This, and the other test converts fear into boredom and I would write no further tests.

Any objections?

olekscode (approved):

PR #256, opened 2022-04-23
Refactor complex number tests: Grouped all arithmetic tests into a clearly named category, added missing interesting test

hemalvarambhia:
- Categorised the tests that exercise the arithmetic of complex numbers;
- renamed some tests to be a little clearer (inspired by Kevlin Henney's [Structure and Interpretation of Test Cases](https://youtu.be/tWn8RA_DEic))
- added what may be a missing test to demonstrate we cannot take the reciprocal of 0 + 0 i

### DIFF
--- a/src/Math-Complex/PMComplex.class.st
+++ b/src/Math-Complex/PMComplex.class.st
@@ -390,7 +390,7 @@ PMComplex >> arcTan: denominator [
 			res]
 ]
 
-{ #category : #arithmetic }
+{ #category : #'polar coordinates' }
 PMComplex >> arg [
 	"Answer the argument of the receiver."
 
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -46,7 +46,7 @@ PMComplexTest >> testAdaptToCollectionAndSend [
 ]
 
 { #category : #'testing - arithmetic' }
-PMComplexTest >> testAddPolynomial [
+PMComplexTest >> testAddingToAPolynomial [
 	| c poly |
 	c := 6 - 6 i.
 	poly := PMPolynomial coefficients: #(1 1 1).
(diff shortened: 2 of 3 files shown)


src/Math-Complex/PMComplex.class.st
```diff
@@ -390,7 +390,7 @@ PMComplex >> arcTan: denominator [
 			res]
 ]
 
-{ #category : #arithmetic }
+{ #category : #'polar coordinates' }
```
hemalvarambhia, line 393: I think this is worth doing because 

`z = r exp(I theta)`

is the polar coordinate form which is taught as a concept explicitly. I feel the code could reflect that too,

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -334,6 +326,14 @@ PMComplexTest >> testDivision1 [
 
 ]
 
+{ #category : #'testing - arithmetic' }
+PMComplexTest >> testDividingPolynomialByAComplexNumber [
```
hemalvarambhia, line 330: Having renamed this, might this be better in the Polynomial tests?

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -775,6 +775,11 @@ PMComplexTest >> testWeCanWriteComplexNumbersWhoseRealAndImaginaryPartsAreFracti
 	self assert: (z imaginary) equals: (Fraction numerator: 4 denominator: 5).
 ]
 
+{ #category : #'testing - arithmetic' }
+PMComplexTest >> testWeCannotTakeReciprocalOfZeroComplexNumbers [
```
hemalvarambhia, line 779: I'll check if this isn't tested somewhere already...

hemalvarambhia:
The idea of the set of complex numbers being a unordered field may mean that we might categorise our arithmetic tests to make that idea explicit.

SergeStinckwich (approved):

PR #261, opened 2022-04-29
Refactor complex tests: Classify tests to enhance their clarity and discover a TestCase break up strategy

hemalvarambhia:
Continued to classify the methods to try to improve test clarity and see if there is a way to break a very large test case up.

### DIFF
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -244,7 +244,7 @@ PMComplexTest >> testComplexCollection [
 
 	complexNumbersMultiplied
 		with: expectedComplexNumbers
-		do: [ :complexNumber :expected | self assert: expected equals: complexNumber ]
+		do: [ :complexNumber :expected | self assert: complexNumber equals: expected ]
 ]
 
 { #category : #'testing - arithmetic' }
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -234,7 +234,9 @@ PMComplexTest >> testComplexCollection [
 
 	| complexNumbers complexNumbersMultiplied |
 	complexNumbers := Array with: 1 + 2 i with: 3 + 4 i with: 5 + 6 i.
+	
 	complexNumbersMultiplied := 2 * complexNumbers.
+	
 	complexNumbers
 		with: complexNumbersMultiplied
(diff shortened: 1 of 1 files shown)


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -230,7 +230,7 @@ PMComplexTest >> testCloseToWithPrecision [
 PMComplexTest >> testComplexAsComplex [
 	| ineg |
 	ineg := -1 i.
-	self assert: ineg asComplex == ineg.
```
nicolas-cellier-aka-nice, line 233: does `equals` equals `equalsequals`?

hemalvarambhia: It's a value object, so I would have thought equality is paramount. I did actually consider removing the test altogether actually

hemalvarambhia: I see your point about style - pharo was complaining hence the commit message

hemalvarambhia: @nicolas-cellier-aka-nice I'd get rid of this test personally. What do you feel?

If we keep it, then we ought not to test identity of value objects. Identity of value objects may be nonsensical.

nicolas-cellier-aka-nice: Hi Hemal, you make a good point about value objects: as long as we don't mutate these kind of object, we don't really care about identity, that's an implementation detail.

If we take this point of view, maybe the point of the test was to check that implementation is optimal: no need to create a new Complex?

Otherwise, the purpose of `asComplex` is to guaranty that the returned object will behave as a complex.
Typically, `-1 asComplex sqrt` could answer `1i`, rather than raising a `DomainError`.

If we just test equality, it's too weak, because we can have `-1 asComplex = -1`, which tells us nothing about how `-1 asComplex` will behave.

Since asComplex is so obvious for complex receiver, I agree with you that the test has very low value anyway if testing identity, my point is that is has even less value if testing equality.

hemalvarambhia: Cool. You've clarified the behaviour well enough that it's worth renaming the test method!

I'd like to remove this test, then.

hemalvarambhia: The point about `-1 sqrt` being the imaginary is a *interesting* enough test to me that it's worth writing.

PR #268, opened 2022-05-03
Refactor complex number tests

hemalvarambhia:
- continued to classify test methods clearly;
- clarified the names of the test to help decide whether to keep them or not.

### DIFF
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -63,7 +63,7 @@ PMComplexTest >> testAddingTwoComplexNumbers [
 	self assert: c equals: 0 + 2 i
 ]
 
-{ #category : #tests }
+{ #category : #'testing - arithmetic' }
 PMComplexTest >> testAdditionIsCommutativeWithAllRealNumbers [
 	self assert: 1 + 2 i + 1 equals: 2 + 2 i.
 	self assert: 1 + (1 + 2 i) equals: 2 + 2 i.
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -298,7 +298,7 @@ PMComplexTest >> testCosh2MinusSinh2 [
 		self assert: ((c cosh squared - c sinh squared) imaginary closeTo: 0.0)]]
 ]
 
-{ #category : #tests }
+{ #category : #'testing - expressing complex numbers' }
 PMComplexTest >> testCreation [
 	"self run: #testCreation"
 
@@ -498,7 +498,7 @@ PMComplexTest >> testNegated [
 	self assert: c negated equals: -2 - 5 i
 ]
 
-{ #category : #tests }
+{ #category : #'testing - expressing complex numbers' }
 PMComplexTest >> testNew [
 	| c |
 	c := PMComplex new.
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -239,10 +239,6 @@ PMComplexTest >> testComplexNumberAndIntegerAreUnequalEvenIfRealPartIsEqualToInt
 
 { #category : #tests }
 PMComplexTest >> testConversion [
-	"self run: #testConversion"
-
-	"self debug: #testConversion"
-
 	self assert: 1 + 2 i + 1 equals: 2 + 2 i.
 	self assert: 1 + (1 + 2 i) equals: 2 + 2 i.
 	self assert: 1 + 2 i + 1.0 equals: 2.0 + 2 i.
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -64,7 +64,7 @@ PMComplexTest >> testAddingTwoComplexNumbers [
 ]
 
 { #category : #tests }
-PMComplexTest >> testAdditionIsCommutativeWithAllNumbers [
+PMComplexTest >> testAdditionIsCommutativeWithAllRealNumbers [
 	self assert: 1 + 2 i + 1 equals: 2 + 2 i.
 	self assert: 1 + (1 + 2 i) equals: 2 + 2 i.
 	
--- a/src/Math-Tests-Complex/PMComplexTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexTest.class.st
@@ -63,6 +63,18 @@ PMComplexTest >> testAddingTwoComplexNumbers [
 	self assert: c equals: 0 + 2 i
 ]
 
+{ #category : #tests }
+PMComplexTest >> testAdditionIsCommutativeWithAllNumbers [
+	self assert: 1 + 2 i + 1 equals: 2 + 2 i.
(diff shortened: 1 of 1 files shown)


src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -63,6 +50,18 @@ PMComplexTest >> testAddingTwoComplexNumbers [
 	self assert: c equals: 0 + 2 i
 ]
 
+{ #category : #'testing - arithmetic' }
+PMComplexTest >> testAdditionIsCommutativeWithAllRealNumbers [
```
hemalvarambhia, line 54: I've wondered whether we should add a test to demonstrate commutativity and associativity of addition. I think it is the definition of a (mathematical) field. Similarly, we might demonstrate the commutativity of multiplication, another part of the definition of a mathematical field.

@olekscode @SergeStinckwich @khinsen what are your thoughts. This could become a contract test for all fields, where addition and multiplication are commutative and associative.

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -833,6 +818,39 @@ PMComplexTest >> testTwoComplexNumbersWithDifferentRealPartsAreNotEqual [
...
+	c := 6 - 6 i.
+	arr := #(0 1 2 3 4 5).
+	
+	expected := Array new: 6.
+	expected at: 1 put: 6 - 6 i.
+	expected at: 2 put: 7 - 6 i.
+	expected at: 3 put: 8 - 6 i.
+	expected at: 4 put: 9 - 6 i.
+	expected at: 5 put: 10 - 6 i.
+	expected at: 6 put: 11 - 6 i.
```
hemalvarambhia, line 833: Could this be written more concisely?

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -237,16 +218,14 @@ PMComplexTest >> testComplexNumberAndIntegerAreUnequalEvenIfRealPartIsEqualToInt
...
-{ #category : #tests }
+{ #category : #'testing - arithmetic' }
 PMComplexTest >> testConversion [
-	"self run: #testConversion"
-
-	"self debug: #testConversion"
-
 	self assert: 1 + 2 i + 1 equals: 2 + 2 i.
 	self assert: 1 + (1 + 2 i) equals: 2 + 2 i.
+
```
hemalvarambhia, line 225: I've wondered whether we might add separate tests to demonstrate **commutativity** and **associativity** of addition. I think it is the definition of a (mathematical) field. Similarly, we might demonstrate the commutativity of multiplication, another part of the definition of a mathematical field.

@olekscode @SergeStinckwich @khinsen what are your thoughts?. This could become a contract test for all fields, where addition and multiplication are commutative and associative.

hemalvarambhia: I'm not sure I find this test useful. Does anyone else feel different?

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -32,19 +32,6 @@ PMComplexTest >> testAbsSecure [
...
-PMComplexTest >> testAdaptToCollectionAndSend [
-	"self run: #testAbsSecure"
-
-	"using equalsTo since absSecure returns a slightly different Float"
-
-	| c arr |
-	c := 6 - 6 i.
-	arr := #(0 1 2 3 4 5).
-	self assert: (arr + c at: 1) equals: c.
-	self assert: (arr * c at: 2) equals: c
```
olekscode, line 45: Yes, this was a very bad test
Good that you removed it

src/Math-Tests-Complex/PMComplexTest.class.st
olekscode: Yes, that could be a good idea
For example:
```st
testAddition
    self assert: 2 + 3 equals: 5.

testAdditionIsCommutative
    self assert: 2 + 3 equals: 3 + 2.
```
And so on

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -237,16 +218,14 @@ PMComplexTest >> testComplexNumberAndIntegerAreUnequalEvenIfRealPartIsEqualToInt
 	self deny: 1 + 3 i equals: 1.
 ]
 
-{ #category : #tests }
+{ #category : #'testing - arithmetic' }
 PMComplexTest >> testConversion [
```
olekscode, line 222: I don't understand why this test is called `testConversion`
It seems to be testing different arithmetical operations

I would split it into multiple tests with meaningful names (one assertion per test)
For example: `testAdditionIntegerAndComplex`, `testAdditionFloatAndComplex`, `testAdditionComplexAndNetagiveFloat`, etc.

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -833,6 +796,38 @@ PMComplexTest >> testTwoComplexNumbersWithDifferentRealPartsAreNotEqual [
  	self deny: z equals: w.
 ]
 
+{ #category : #'testing - arithmetic' }
+PMComplexTest >> testWeCanAddComplexNumbersToAnArrayOfIntegers [
```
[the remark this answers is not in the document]
olekscode: I don't know, but this seems like a good test to have
And I like the name
Maybe make it shorter: `testAddComplexToArrayOfIntegers`

I think this `WeCan` is never a good idea because for any test, if it is passing, then `yes we can!` :)
Haha

src/Math-Tests-Complex/PMComplexTest.class.st
olekscode: For sure!
```st
expected := { 6-6i . 7-6i . 8-6i . 9-6i . 10-6i . 11-6i }.
```

src/Math-Tests-Complex/PMComplexTest.class.st
```diff
@@ -833,6 +796,38 @@ PMComplexTest >> testTwoComplexNumbersWithDifferentRealPartsAreNotEqual [
...
+	c := 6 - 6 i.
+	arrayOfIntegers := #(0 1 2 3 4 5).
+	
+	expected := Array new: 6.
+	expected at: 1 put: 0 + 0 i.
+	expected at: 2 put: 6 - 6 i.
+	expected at: 3 put: 12 - 12 i.
+	expected at: 4 put: 18 - 18 i.
+	expected at: 5 put: 24 - 24 i.
+	expected at: 6 put: 30 - 30 i.
```
olekscode, line 827: And same here

olekscode (approved):

src/Math-Tests-Complex/PMComplexTest.class.st
hemalvarambhia: That's what I thought!

I would have preferred some close to what you suggested. They are testing addition to Real Numbers, Integers and Rational Numbers.

All these makes me wonder about our tests and making them more domain-driven and documentation like.

src/Math-Tests-Complex/PMComplexTest.class.st
hemalvarambhia: Almost, I split it up, and chose a Conservative route, later seeking support from you and Serge on removing it.

PR #270, opened 2022-05-03
Fixed #265. Moved asPMVector from Math-Numerical to Math-Core

### LINKED ISSUE #265: Move Collection >> asPMVector to Math-Core
It should be together with `PMVector` itself and not in `Math-Numerical`

olekscode:


hemalvarambhia (approved):
Another alternative might be Linear Algebra.

olekscode:
@hemalvarambhia what do you mean? A package name?
I don't know... we have too many things that could go in that package
For now, I just moved the method `asPMVector` into the same package where `PMVector` is

Later, I want to separate `PMVector` and `PMMatrix` into `Math-Collections`
Because every other package (Algebra, Statistics, Differential Equations, etc.) depends on those collections

But we can discuss it later

PR #271, opened 2022-05-03
Fixed #266. Moved productWithVector: and productWithMatrix: closer to PMVector and PMMatrix

### LINKED ISSUE #266: Move Number >> productWithVector: and productWithMatrix: from Math-Numerical to Math-Core and Math-Matrix

olekscode:


src/Math-Core/Number.extension.st
```diff
@@ -11,3 +11,9 @@ Number >> addWithVector: aVector [
 	"Adds itself to each element of the vector"
 	^ aVector collect: [ :each | each + self ]
 ]
+
+{ #category : #'*Math-Core' }
+Number >> productWithVector: aVector [
```
hemalvarambhia, line 16: This represents a scalar multiplied by a vector?

Perhaps `timesVector: aVector`

src/Math-Matrix/Number.extension.st
```diff
@@ -0,0 +1,6 @@
+Extension { #name : #Number }
+
+{ #category : #'*Math-Matrix' }
+Number >> productWithMatrix: aMatrix [
```
hemalvarambhia, line 4: `timesMatrix: aMatrix` perhaps?

hemalvarambhia (approved):
Small non-blocking suggestions.

olekscode:
@hemalvarambhia you are right, those would be good names if the methods were part of public API
But those are private methods that are used for double dispatch:
```st
PMvector >> * anObject
    ^ anObject productWithVector: self
```
```st
Number >> productWithVector: aVector
    ...
    
    
PMMatrix >> productWithVector: aVector
    ...
```
So it can not be `timesVector:`, otherwise we'll have `matrix timesVector: vector` :)

PR #272, opened 2022-05-03
Fixed #267. Deprecated equalsTo: and changed all senders to use closeTo: instead

### LINKED ISSUE #267: Remove equalsTo: because it is closeTo: and not equalsTo:
This is the implementation of `Number >> equalsTo:` defined in `Math-Numerical`:
```st
Number >> equalsTo: aNumber
		"compare to Float>>closeTo:"
	^self relativelyEqualsTo: aNumber upTo: PMFloatingPointMachine new defaultNumericalPrecision
```
This is same as `closeTo:` which is now understood by any Number and not just Float

Also, this is causing some confusion, because we are not checking equality but "closeness".
So I propose to remove this method and replace all references with `closeTo:`

olekscode:
Fixing issue #267

### DIFF
--- a/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
+++ b/src/Math-AutomaticDifferenciation/PMDualNumber.class.st
@@ -133,6 +133,13 @@ PMDualNumber >> asInteger [
 	^ value asInteger
 ]
 
+{ #category : #comparing }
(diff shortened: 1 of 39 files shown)


hemalvarambhia (approved):
:heart:

Issue #191, opened 2021-02-23
Rename PMComplex to PMComplexNumber


olekscode:

SergeStinckwich:
I'm not sure we should go in that direction. For example, there is a ``Float`` or ``Integer`` not ``FloatNumber`` or ``IntegerNumber``.

nicolas-cellier-aka-nice:
The main grief was with vocable `isComplex` which is way too broad and subject to false polymorphism.
A lot of objects can be complex, without being complex numbers.
In VW, there was at least a case of such false polymorphism (in ProgramNode).
Same in Squeak (ParseNode) and (ComplexBorder).

So the initial proposal was to rename `isComplex` to `isComplexNumber`.
I think that the class renaming idea originated from this pattern: `Integer >> #isInteger`, `Float >> #isFloat`, etc...
But with the PM prefix, maybe the renaming is moot.
Unless we rename `isComplex` -> `isPMComplex`...
Beware: `isPMComplex` might reduce possibilities of extensions/intercompatibility because on the contrary to `isComplex`, it could be seen as much too restrictive. Some other species might quake like a complex...

olekscode:
Well, to be honest, i also think that in a general scenario, `Integer` and `Float` should be `IntegerNumber` and `FloatNumber`.

But because `int` and `float` are super-standard terms in every programmer's vocabulary, as an exception, we can keep those class names simple.

But `Complex` is a different story. It is an adjective (class names should be nouns!) that is often used in various contexts: complex view, complex scenario, etc. This can confuse many programmers who read the code, especially those who are not mathematicians.

In other words, this:

```Smalltalk
x := PMComplex new.
x isComplex.
x asComplex.
```

seems to me much less descriptive than this:

```Smalltalk
x := PMComplexNumber new.
x isComplexNumber.
x asComplexNumber.
```

hemalvarambhia:
I would like weigh in, if I may. I'm wondering about this from a domain driven  perspective. I think it should be PMComplexNumber, and Integer remain the same. This rename makes sense to me with my background, as we already referred to 'integers' and 'complex numbers' in conversation. We should refactor to what is proposed here.

There is a wider question of the project appealing to Domain Driven Design. I think it should. Is there any reason we shouldn't adopt Domain Driven Design (DDD)? At the very least, let's have the code be as close as possible to the domain language of mathematics. 

[...]

PR #276, opened 2022-05-04
Breaking Change: Renamed the class to follow the ubiquitous language.

hemalvarambhia:
Issue: #191 
**Breaking Change**

- Renamed the PMComplex class and its Test Case to PMComplexNumber and PMComplexNumberTest respectively;
- migrated clients.

I have chosen to rename the package as a separate PR, as this step has proved difficult.


olekscode (approved):
Yes! I think this is a great idea

hemalvarambhia:
@SergeStinckwich @olekscode this is a breaking change and would require a bump to the major version number. Would you prefer I add a new Complex Number class and migrate clients instead?

PR #287, opened 2022-07-31
Refactor: Extract QR Decomposition Class

hemalvarambhia:
This is to apply the Replace Method With Method Object Refactoring as a starting point.

QR Decomposition is an operation that mathematicians refer to, so it Is worth making explicit in the form of a class

### DIFF
--- a/src/Math-Matrix/PMQRDecomposition.class.st
+++ b/src/Math-Matrix/PMQRDecomposition.class.st
@@ -19,6 +19,12 @@ PMQRDecomposition class >> of: matrix [
 ]
 
 { #category : #private }
+PMQRDecomposition >> columnVectorFrom: r startingAt: col [
+
+	^ (r columnAt: col) copyFrom: col to: colSize
+]
+
+{ #category : #private }
 PMQRDecomposition >> decompose [
 
 	| identityMatrix q r householderVector i |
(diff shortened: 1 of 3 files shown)


src/Math-Matrix/PMQRDecomposition.class.st
```diff
@@ -0,0 +1,91 @@
...
+
+{ #category : #'instance creation' }
+PMQRDecomposition class >> of: matrix [
+	matrix numberOfRows < matrix numberOfColumns ifTrue: [ 
+		self error: 'numberOfRows<numberOfColumns' ].
+	^ self new of: matrix
+]
+
+{ #category : #private }
+PMQRDecomposition >> columnVectorFrom: r startingAt: col [
```
hemalvarambhia, line 22: This method could be inlined as it is obsessed with `r`.

SergeStinckwich:
Thank you Hemal for spending time to decompose complex methods in less complex ones. This is an important task to do that. I'm not sure to understand why you need a separate class. Do you expect this ``PMQRDecomposition`` class to be used in other places?

hemalvarambhia:
Possibly bur that was not the main driver. This is a standard technique for dealing with long methods, and I was seeing the need to compute state that does not belong with a matrix.

PR #289, opened 2022-09-09
Refactor QR Decomposition

hemalvarambhia:
This continues the simplification of the QR Decomposition method object. I have made the idea of the Householder Reflection explicit. I will continue to look more deeply into why the one element of the pivot (last one) is nil.


src/Math-Matrix/PMQRDecomposition.class.st
```diff
@@ -29,64 +29,67 @@ that describes the mechanics:
...
-		v := (PMVector zeros: col - 1) , (householderVector at: 2).
-		householderMatrix := identityMatrix
-		                     -
-		                     ((householderVector at: 1) * v tensorProduct: v).
+		householderReflection := self
+			                         householderReflectionOf:
+			                         columnVectorFromRMatrix
+			                         atColumnNumber: col.
+		householderVector := householderReflection at: 1.
+		householderMatrix := householderReflection at: 2.
```
hemalvarambhia, line 41: I want Housholder Reflection to become a first class object.

jordanmontt (approved):

PR #292, opened 2022-11-28
Cut dependency from Core to Complex.

jecisc:
With this, Math-Core does not depend on anything else from Polymath.


jecisc:
image

olekscode (approved):

PR #309, opened 2023-02-23
Launch tests of all packages starting with Math-

jecisc:
This will allow to launch tests of project that are externalized


jecisc:
I also simplified a CI file + changed two things to remove deprecation warnings

olekscode (approved):

PR #315, opened 2023-03-09
Use pharo precision

jecisc:
Try to use Pharo precision to see if polymath can run with this one


jecisc:
This breaks lot of stuff :/

SergeStinckwich:
Thank you Cyril for the report. I think we can keep it like that for the moment and we can come back to this issue later, when a modular version of PM.

PR #321, opened 2023-05-25
Updated examples to plot PCA with Roassal 3

hernanmd:
This needs https://github.com/PolyMathOrg/DataFrame/pull/237 to be merged.
Roassal


jecisc:
Polymath has no dependency on Roassal and I don't think it is a good idea to depend on a visualization framework for a math framework. 

Maybe @olekscode had an opinion?

jordanmontt:
I agree with Cyril, we should not put a dependency to Roassal

hernanmd:
Could we have a Baseline group for loading visualizations examples as extension methods and so on?

jordanmontt:
In my opinion, this should be in the [data-inspector](https://github.com/pharo-ai/data-inspector) repo since we have there other visualizations as well

SergeStinckwich:
Can we move all PolyMath visualisations in the PolyMath org?

jordanmontt:
I'm not against creating a new repo inside PolyMath for visualizations. But, we should be careful to not have duplicated visualizations with [data-inspector](https://github.com/pharo-ai/data-inspector).
We should discuss this. Maybe in the next ai meeting?

jecisc:
I'm also not against it. For me data-inspectors is really about visualizations to help take decisions about your data. If we do on for PolyMath, IMA it should be things helping to visualize Polymath related things. Not visualizations to help take decisions over datas

hernanmd:
I repackaged it in a new package `Math-Visualizations` and updated the Baseline references.
It also contains a new group `visualizations` to load PolyMath code related with graphics.

Ducasse:
You have two approach
- either another project PMWithVisualization with dependency to roassal and PM
- or a package and you factor the dependency: you can load some PM packages without the dependency (but the dependency is still at the level of project).

hernanmd:
Yes, this PR fixes the dependency problem. If you load it normally PM packages are loaded without any visualization dependency:

```smalltalk
Metacello new
        repository: 'github://PolyMathOrg/PolyMath:plot_pca';
        baseline: 'PolyMath';
        load
```

So there is no Roassal dependency (except a Roassal reference from another package, "Math-TSNE", which I will address in another PR).

If you load the branch of this PR, there will be a new package `Math-Visualizations`:

```smalltalk
Metacello new
	repository: 'github://PolyMathOrg/PolyMath:plot_pca';
	baseline: 'PolyMath';
	load: #('visualizations')
```

I tested under Pharo 11 and Pharo 12 and it works as expected.

Screenshot 2023-12-28 at 20 00 21

It is already safe to be integrated

jordanmontt:
It looks good for me

jordanmontt (approved):

hemalvarambhia (approved):

PR #322, opened 2023-12-24
Refactor Polynomials: Extract Method to Introduce RootFinder Factory Method and Tighten Encapsulation

hemalvarambhia:
- Extracted a factory method that generates a root finder with a given precision;
- added some missing tests that felt interesting.
Another test might be to test roots that are in the Complex plane.

### DIFF
--- a/src/Math-Polynomials/PMPolynomial.class.st
+++ b/src/Math-Polynomials/PMPolynomial.class.st
@@ -240,6 +240,15 @@ PMPolynomial >> reciprocal [
 ]
 
 { #category : #information }
+PMPolynomial >> rootFindingAlgorithmWith: precision [
+
+	| rootFinder |
+	rootFinder := PMNewtonZeroFinder new.
+	rootFinder desiredPrecision: precision.
+	^ rootFinder
+]
+
+{ #category : #information }
 PMPolynomial >> roots [
 
 	^ self roots: Float defaultComparisonPrecision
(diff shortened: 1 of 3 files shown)


src/Math-Tests-Polynomials/PMPolynomialTest.class.st
```diff
@@ -235,13 +235,36 @@ PMPolynomialTest >> testPolynomialRoots [
...
-	roots := polynomial roots asSortedCollection asArray.
+	roots := polynomial roots .
 	self assert: roots size equals: 3.
 	self assert: (roots at: 1) + 2 closeTo: 0.
 	self assert: (roots at: 2) + 1 closeTo: 0.
 	self assert: (roots at: 3) - 5 closeTo: 0
 ]
 
+{ #category : #'iterative algorithms' }
+PMPolynomialTest >> testPolynomialRootsForConstant [
```
hemalvarambhia, line 246:
```suggestion
PMPolynomialTest >> testPolynomialRootsConstantsHaveNoRoots [
```

src/Math-Polynomials/PMPolynomial.class.st
```diff
@@ -26,6 +26,15 @@ PMPolynomial class >> coefficients: anArray [
 	^self new initialize: anArray reverse
 ]
 
+{ #category : #information }
```
Ducasse: I'm not sure I would put this method on the class side but I could understand. 
Now why the PMNewtonZeroFinder does not define a creation class method  desiredPrecision: like that you do not need this method and the class PMNewtonZeroFinder gets improved.

Ducasse: I saw that you did it :)

hemalvarambhia: Yes, I had to take smaller steps in order to keep our tests green.

PR #323, opened 2023-12-27
Refactor polynomials

hemalvarambhia:
- Improved naming;
- Tightened Encapsulation;
- Added some missing tests, including demonstrating that multiplication is commutative.
- Made tests more like examples written with DSLs

### DIFF
--- a/src/Math-Polynomials/PMPolynomial.class.st
+++ b/src/Math-Polynomials/PMPolynomial.class.st
@@ -245,10 +245,7 @@ PMPolynomial >> roots: aNumber [
 		       (coefficients reverse collect: [ :each | each asFloat ]).
 	roots := OrderedCollection new: self degree.
 	[
-	rootFinder
-		setFunction: pol;
-		setDerivative: pol derivative.
-	root := rootFinder evaluate.
+	root := rootFinder rootOf: pol.
 	rootFinder hasConverged ] whileTrue: [
 		roots add: root.
 		pol := pol deflatedAt: root.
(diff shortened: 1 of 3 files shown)


hernanmd:
There is only 1 test failing `testOptimize` on linux

hemalvarambhia:
> There is only 1 test failing `testOptimize` on linux
> 

Restarted the CI run.

src/Math-Polynomials/PMPolynomial.class.st
```diff
@@ -239,20 +239,17 @@ PMPolynomial >> roots [
 { #category : #information }
 PMPolynomial >> roots: aNumber [
 
-	| pol roots root rootFinder |
+	| polynomial roots root rootFinder |
 	rootFinder := PMNewtonZeroFinder with: aNumber.
```
hemalvarambhia, line 243: Is `aNumber` the starting point for computing the root i.e. 
$`x_0`$

hernanmd: Is that a question?

hemalvarambhia: Yes. It is a question.

hernanmd: aNumber is the initial guess for the root (in PMNewtonZeroFinder x0 is the initial guess for the root, which is passed to the `initValue:` method, which sets the initial value for the iterations)

hernanmd (approved):

PR #325, opened 2024-01-11
Add tests for complex numbers

hemalvarambhia:
- Added some missing tests

### DIFF
--- a/src/Math-Tests-Complex/PMComplexNumberTest.class.st
+++ b/src/Math-Tests-Complex/PMComplexNumberTest.class.st
@@ -254,6 +254,14 @@ PMComplexNumberTest >> testComplexNumberAndIntegerAreUnequalEvenIfRealPartIsEqua
 ]
 
 { #category : #'testing - mathematical functions' }
+PMComplexNumberTest >> testComplexNumberZeroRaisedToThePowerOfZero [
+
+	| zeroRaisedToZero |
+	zeroRaisedToZero := (PMComplexNumber zero) raisedTo: 0.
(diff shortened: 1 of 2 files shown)


src/Math-Tests-Complex/PMComplexNumberTest.class.st
```diff
@@ -637,6 +637,15 @@ PMComplexNumberTest >> testRaisedTo [
 	self assert: c3 imaginary closeTo: c imaginary
 ]
 
+{ #category : #'testing - mathematical functions' }
+PMComplexNumberTest >> testRaisedToFractionalPower [
+
+	| z expected |
+	z := 0 + 1 i.
```
hemalvarambhia, line 644: When $0 + 1 i = cos(\pi/2) + sin(\pi/2)i$, its cube root is $cos(\pi/6) + i sin(\pi/6) = \sqrt(3)/2 + 1/2i$ by De Moivre's theorem

hernanmd: Looks good to me. For better clarification while the variable name z is commonly used for complex numbers, consider using a more descriptive name that conveys its purpose in the test. For example, you could use complexNumber instead.

hernanmd (approved):

jordanmontt (approved):

hemalvarambhia:
@hernanmd yes. I use it because it feels natural (I did a Theoretical Physics degree and z or w were used all the time) it makes it so much more DSL like

PR #332, opened 2026-03-18
Adjustments following changes made to JacobiTransformation

DurieuxPol:


DurieuxPol:
Alright, I checked each Pharo version locally, and the failing tests don't seem to be related to my changes and they failed already in previous versions:
- `PMNumericalMethodsTestCase>>#testOptimize` is random from what I saw, it usually passes but sometimes it fails ?
- `PMQRTest>>#testDecompositionOfMatrixCausingErraticFailure` is expected to fail but passes in P11 and P12 for some reason
- In `PMClusterFinderTest`, `#testClusterCovariance` and `#testClusterEuclidean` fails for all except P11

So over all I think it's good to merge

PR #333, opened 2026-03-25
Fixed remaining red tests

DurieuxPol:
Several `testPrint` tests of FunctionFit were broken because the way BlockClosure are printed as strings changed in recent Pharo versions, so I adapted the code around it.  
Also, the way Random is initialised with seed has changed as well in recent Pharo versions, so I changed the seed in the setup of `PMClusterFinderTest` so that the tests are green in all versions.


DurieuxPol:
The last red test is `PMQRTest>>testDecompositionOfMatrixCausingErraticFailure` mentioned also in issue 10 of vector-matrix. I think it needs more work to solve the issue, so I will do another PR for that.  
So this one is good to merge !

